PR #13, opened 2016-06-22
Fixes #10: Render progress bar programmatically.

### LINKED ISSUE #10: Change the headers/payload
There should be a synchronous API to change the payload if we want to, something like

```js
const autocannon = require('autocannon')

const instance = autocannon({
    url: 'http://localhost:3000',
    connections: 1000,
    duration: 20
}, console.log)

instance.on('response', function (client, statusCode, returnBytes, responseTime) {
  client.setHeaders({ .. })
  client.body = ...
}) 
```

thekemkid:
Users can pass `renderProgressBar: true` to render the
progress bar from an external tool. Additionally, they
can provide `progressBarString: SOME_STRING` to modify
the output format of the progress bar. This string must
meet the format specified in npm.im/progress

This will require a review.

Additionally... how should I test this?

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,8 +82,36 @@ Start autocannon against the given target, options:
 * `duration`: the number of seconds to run the autocannon
 * `body`: a `Buffer` containing the body of the request
 * `method`: the http method to use, `GET` is the default
-* `renderProgressBar`: A truthy value to enable the rendering of the progress bar. false by default
-* `progressBarString`: A `string` defining the format of the progress display output. Must be valid input for the [progress bar module](http://npm.im/progress).
+
+**Returns** an instance/event emitter for tracking progress, etc.
+
+### autocannon.track(instance[, outputStream])
+
+* `instance`: the instance of autocannon
+* `opts`: OPTIONAL. configuration options for tracking. this can have the
+following attibutes
+    * `outputStream`: the stream to output to. Defaults to process.stderr
+    * `renderProgressBar`: A truthy value to enable the rendering of the progress bar. true by default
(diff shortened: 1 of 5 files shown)


mcollina:
Thinking twice about this, I think we should just move all the output logic from the main `autocannon.js` to its own module, rather than instantiating it twice. That include also the table.

Something like:

```js
const autocannon = require('autocannon') 
const instance = autocannon({ url: 'http://localhost:3000' }, done)
// customize your autocannon instance here
autocannon.track(instance, process.stdout)
```

Does it make sense? That do you think?

thekemkid:
That sounds okay, I will update my PR to reflect this :)

README.md
```diff
@@ -82,8 +82,36 @@ Start autocannon against the given target, options:
 * `duration`: the number of seconds to run the autocannon
 * `body`: a `Buffer` containing the body of the request
 * `method`: the http method to use, `GET` is the default
-* `renderProgressBar`: A truthy value to enable the rendering of the progress bar. false by default
-* `progressBarString`: A `string` defining the format of the progress display output. Must be valid input for the [progress bar module](http://npm.im/progress).
+
+**Returns** an instance/event emitter for tracking progress, etc.
+
```
mcollina: We should probably list the events of this instance.

thekemkid: I would like to merge both of my open PR's and then do docs on the events emitted by the instance, as the 'response' event needs to be explained, and the client object which it passes to the cb: https://github.com/mcollina/autocannon/pull/14/files#diff-de78b50723ff596d7bd35665f36a6cbfR60

mcollina:
This one is good to go then :).

PR #14, opened 2016-06-23
fixes #9: Can now change headers and body on client response

### LINKED ISSUE #9: Export the progress bar and reporting
The goal is to let users build their own tool if they want to, embedding autocannon.

thekemkid:
You need to call `.setHeaders(newHeaders)` and `.setBody(newBody)` for modifying these values. The reason you can't shouldn't just assign `.body` is because they are used to build a cached request, and simply overwriting it would mean I would have to un-cache the request and build it every time the client makes the request, because wouldn't be able to effectively and efficiently detect when it has been changed... simply put, its a big no-no for perf. 😄

If you do need to assign directly to the body though, it is possible, you just need to rebuild your request like so:

```js
const instance = autocannon({
    url: url,
    connections: 1000,
    duration: 10
  }, finishedBench)

  instance.on('response', function (client, statusCode, returnBytes, responseTime) {
    client.setHeaders({ .. })
    client.body = ...
    client.rebuildRequest()
  })
```

However, the preferred method of use is:
```js
const instance = autocannon({
    url: url,
    connections: 1000,
    duration: 10
  }, finishedBench)

  instance.on('response', function (client, statusCode, returnBytes, responseTime) {
    client.setHeaders(...)
    client.setBody(...)
  })
```

Also, `.headers` cant be reassigned to, as I'm not what the `.headers` variable are for.

### DIFF
--- a/lib/myhttp.js
+++ b/lib/myhttp.js
@@ -105,6 +105,7 @@ Client.prototype.setBody = function (newBody) {
 }
 
 Client.prototype.setHeadersAndBody = function (newHeaders, newBody) {
+  this.opts.headers = newHeaders || {}
   this.opts.body = newBody
 
   this._rebuild()
--- a/lib/myhttp.js
+++ b/lib/myhttp.js
@@ -29,7 +29,6 @@ function Client (opts) {
   this.bytes = 0
   this.headers = {}
   this.startTime = [0, 0]
-  this.body = this.opts.body
 
   this.parser[HTTPParser.kOnHeadersComplete] = (opts) => {
     this.headers = opts
@@ -52,7 +51,7 @@ function Client (opts) {
     opts.host = opts.hostname + ':' + opts.port
   }
 
-  this._req = `${opts.method} ${opts.path} HTTP/1.1\r\nHost: ${opts.host}\r\nConnection: keep-alive\r\n`
+  this.baseReq = `${opts.method} ${opts.path} HTTP/1.1\r\nHost: ${opts.host}\r\nConnection: keep-alive\r\n`
 
   this.setHeaders(opts.headers)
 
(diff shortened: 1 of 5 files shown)


lib/myhttp.js
```diff
@@ -29,6 +29,7 @@ function Client (opts) {
   this.bytes = 0
   this.headers = {}
   this.startTime = [0, 0]
+  this.body = this.opts.body
```
mcollina: I think you can avoid this.

lib/run.js
```diff
@@ -48,16 +50,19 @@ function run (opts, cb) {
...
     clients.push(client)
   }
 
-  function record (statusCode, resBytes, responseTime) {
-    statusCodes[(parseInt(statusCode) / 100) - 1] += 1
-    latencies.record(responseTime)
-    bytes += resBytes
-    counter++
+  function record (client) {
+    return function record (statusCode, resBytes, responseTime) {
```
mcollina: I think you can have two different names for this, `buildRecord` and `record` maybe.

mcollina: you don't need to do this, because `this` will be set to the `client` instance

mcollina:
Unit tests are missing.

lib/myhttp.js
thekemkid: I'm not sure, as this will allow the user to assign to `client.body` and when the `rebuildRequest()` is called, it should pick up their assignments. However, I would prefer to take this out and just use `.setBody(...)` so its is a consistent api for headers and body.

mcollina: I agree with consistency. We should probably rename `rebuildRequest` to `_rebuild()`.

lib/run.js
thekemkid: Does the .`on` callback bind to the event emitter? TIL 😄

lib/myhttp.js
```diff
@@ -117,4 +94,45 @@ Client.prototype.destroy = function () {
...
+    throw new Error('body must be either a string or a buffer')
+  }
+
+  if (bodyReq) {
+    this.opts.headers['Content-Length'] = '' + bodyReq.length
+  }
+
+  this._req = Object.keys(this.opts.headers)
+    .map((key) => `${key}: ${this.opts.headers[key]}\r\n`)
+    .reduce((acc, str) => acc + str, this._req)
```
mcollina: This might be a perf issues.

mcollina:
Can you also add an example (maybe in the examples folder) where you are using this API?

lib/run.js
mcollina: yes it does. It's actually called with `call`.

lib/myhttp.js
thekemkid: Well, the request is built/cached once, unless users go crazy with the `.setHeaders()` and `.setBody()` call. If you this improved though, I'll need some guidance :)

mcollina: Let's talk about this later on :)

PR #17, opened 2016-06-24
Issue 15 fix, can now track latency of every pipelined request :)

thekemkid:


README.md
```diff
@@ -76,7 +76,7 @@ autocannon({
 Start autocannon against the given target, options:
 
 * `url`: the given target, mandatory
-* `connections`: the number of concurrent connections
+* `connections`: the number of concurrent connections, mandatory
```
davidmarkclements: how come mandatory, why no default?

mcollina: Do you think that 100 is a good value?

mcollina: It was my bad :/

thekemkid: You use 10 in the readme examples, I think using that is a good idea as it gives a clue to users about what to expect?

thekemkid:
I'm going to close this as my new PR has this included within it.

PR #20, opened 2016-06-24
update docs and defaults

### LINKED ISSUE #19: Documentation and validation work
Currently, the documentation doesn't show how to customise the headers on the request.

For the duration option, browsing the code, it is required, but never checked, @mcollina, should we put a check in for the duration and test it, or default it to some value, eg. `10`? Obviously, this will need a unit test and documentation work, too.

Supported HTTP methods should be documented.

Show the default value for the http pipelining value.

Update the `.track` doc, I left `outputStream` listed as a param name when I should have listed `opts`.

thekemkid:
I will need my other PR to be merged to finish the work on defaults, as there is clashing details between this PR and my other

Fixes #19

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,39 +63,44 @@ const autocannon = require('autocannon')
 
 autocannon({
   url: 'http://localhost:3000',
-  connections: 10,
+  connections: 10, //default
   pipelining: 1, // default
-  duration: 10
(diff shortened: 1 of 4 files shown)


mcollina:
Thx!

PR #22, opened 2016-06-24
Fixed #16: can now pass customiseRequest to build custom requests per connection

### LINKED ISSUE #16: Support a way to set headers and body before the client is started
Maybe we can pass in a function that returns an object, or a function that accepts a client.

What do you think @thekemkid?

### LINKED ISSUE #19: Documentation and validation work

thekemkid:
This has the code, the docs and a sample. No unit tests right now however, I'm unsure how to test this.

fixes #16

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,6 +83,7 @@ Start autocannon against the given target.
     * `method`: The http method to use. _OPTIONAL_ `default: 'GET'`.
     * `body`: A `String` or a `Buffer` containing the body of the request. Leave undefined for an empty body. _OPTIONAL_ default: `undefined`.
     * `headers`: An `Object` containing the headers of the request. _OPTIONAL_ default: `{}`.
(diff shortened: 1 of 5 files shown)


mcollina:
You add a test into myhttp in which you use this to set a body or a header, and you check the server that those were set correctly.

thekemkid:
Cool, I'll add that 👍

thekemkid:
I didn't check the server, just verified that it overwrote the `headers` and `body` options passed in 👍

PR #24, opened 2016-06-24
Fixed #21: added a bailout option.

### LINKED ISSUE #21: Add a bailout option
If running a test against a server which dies, maybe you want your test to stop?

In this case, it might be useful to allow a user to specify a threshold of errors within a specific timeframe to cause a bailout. This would surely damage perf though A better way of managing this is to just exit after the error threshold is hit (not needing an error timeframe).

Examples:

Say you have a server you are benchmarking for half an hour. The server dies after 10 minutes. To detect this, we could either:
Detect if there was 1000 errors from the server within the last second OR we could detect if there was 10000 errors overall.

thekemkid:
Some docs were updated too. The command line option for body was changed, as I thought `-H` for headers and `-B` for the body was a nice piece of consistency.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,10 +47,10 @@ Available options:
   -p/--pipelining NUM   The number of pipelined requests to use
   -d/--duration SEC     The number of seconds to run the autocannnon
   -m/--method METHOD    The http method to use
-  -B/--body BODY        The body of the request
+  -b/--body BODY        The body of the request
   -i/--input FILE       The body of the request
   -H/--headers K=V      The request headers
-  -b/--bailout NUM      The number of failures before initiating a bailout
+  -B/--bailout NUM      The number of failures before initiating a bailout
   -j/--json             Print the output as json
   -l/--latency          Print all the latency data
   -h/--help             Print this menu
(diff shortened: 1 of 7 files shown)


lib/run.js
```diff
@@ -82,6 +87,11 @@ function run (opts, cb) {
 
   function onError () {
     errors++
+    if (opts.bailout) bailout()
```
mcollina: you can avoid this if and move the bailour logic in here.
Set a default for bailout to the max integer.

README.md
```diff
@@ -47,8 +47,10 @@ Available options:
   -p/--pipelining NUM   The number of pipelined requests to use
   -d/--duration SEC     The number of seconds to run the autocannnon
   -m/--method METHOD    The http method to use
-  -b/--body FILE        The body of the request
-  -h/--headers K=V      The request headers
+  -B/--body BODY        The body of the request
+  -i/--input FILE       The body of the request
+  -H/--headers K=V      The request headers
+  -b/--bailout NUM      The number of failures before initiating a bailout
```
mcollina: Please, let's not change option switch

mcollina:
A test is missing

README.md
thekemkid: so -B for bailout?

mcollina: Yes

thekemkid:
wait, I'll cause a bailout by entering an invalid url, and ensure it happened before the duration

PR #29, opened 2016-06-27
added timeout feature

### LINKED ISSUE #23: Timeouts for valid requests from a living server/connection
How are these handled and detected?

thekemkid:
fixes #23 

This will reset a connection when a timeout occurs, because that seemed like the proper approach to clearing all pipelined reqs


mcollina:
is this slowing things down if the timeout is lower than the max duration?

thekemkid:
No, I did a benchmark run with the default timeout (10) and a duration of 30, and it was not slowed down

PR #31, opened 2016-06-27
Fix loop keeping event loop alive

thekemkid:


lib/myhttp.js
```diff
@@ -42,14 +42,20 @@ function Client (opts) {
...
     this.cer = 0
     this._destroyConnection()
     this._connect()
-  }, this.timeout)
+
+    // timeout has already occured, need to set a new timeoutTicker
+    this.timeoutTicker = retimer(handleTimeout.bind(this), this.timeout)
+  }
+
+  this.timeoutTicker = retimer(handleTimeout.bind(this), this.timeout)
```
mcollina: Never use `.bind()` it creates slow functions.

lib/myhttp.js
```diff
@@ -42,14 +42,20 @@ function Client (opts) {
...
+  this.destroyed = false
 
-  this.timeoutTicker = retimer(() => {
+  const handleTimeout = function () {
     // all pipelined requests have timed out here
     this.resData.forEach(() => this.emit('timeout'))
     this.cer = 0
     this._destroyConnection()
     this._connect()
-  }, this.timeout)
```
mcollina: either use a named function and `that`, or an arrow function.

lib/run.js
```diff
@@ -141,6 +141,11 @@ function run (opts, cb) {
         'non2xx': statusCodes[0] + statusCodes[2] + statusCodes[3] + statusCodes[4]
       }
       tracker.emit('done', result)
+
+      tracker.removeAllListeners('done')
+      tracker.removeAllListeners('tick')
+      tracker.removeAllListeners('response')
```
mcollina: why?

PR #34, opened 2016-06-29
updated hdr dependency

thekemkid:


mcollina:
let's wait when appveyor and travis finishes before releasing.

PR #36, opened 2016-06-30
begin refactoring to allow passing array of requests

thekemkid:


mcollina:
I would make the buildRequest a function wrapping the opts in

mcollina:
Sorry about that, I'm on mobile and I pushed the wrong button :/.

Forget about that comment, is perf staying the same?

mcollina:
I would like the builder to have state, meaning it's the builder that maintains the request to be sent.

thekemkid:
Perf is the same :) I'll modify and make this work as appropriate

mcollina:
unit tests? :)

lib/httpClient.js
```diff
@@ -130,24 +121,6 @@ Client.prototype.destroy = function () {
...
-  request.body = newBody || new Buffer(0)
-
-  this._rebuild()
-}
-
-Client.prototype.setRequestHeadersAndBody = function (request, newHeaders, newBody) {
-  request.headers = newHeaders || {}
-  request.body = newBody || new Buffer(0)
-
-  this._rebuild()
```
mcollina: You can call the iterator from these methods.

lib/run.js
```diff
@@ -5,6 +5,18 @@ const URL = require('url')
...
+
+const defaultOptions = {
+  headers: {},
+  body: new Buffer(0),
+  method: 'GET',
+  duration: 10,
+  connections: 10,
+  pipelining: 1,
+  timeout: 10,
+  requests: [{}]
```
mcollina: I don't think that this default is correct, `null` might be better.

lib/httpClient.js
thekemkid: I actually started that code, but have a failing test. Will put it in :)

lib/run.js
thekemkid: I don't think it is, if the users just pass headers and a body I build it into a single request object in the array, which makes the requests collection usable for a single or multiple requests

thekemkid:
I need unit tests and documentation! I do plan on working on it :D

mcollina:
Great, good work!

PR #44, opened 2016-07-05
Title json

### LINKED ISSUE #43: add a --name (or --title) swith
The name (or title) switch should be emitted in the reaults. We may even add the time at which it finished. This makes the json results easier to debug.

thekemkid:
fixes #43 

Also found an issue where the timeout flag was missing.


mcollina:
You know I'm asking you to add a unit test :)

thekemkid:
I added some assertions, not sure how else I could test that? Are those okay? :)

lib/run.js
```diff
@@ -139,12 +139,15 @@ function run (opts, cb) {
...
       let result = {
+        title: opts.title,
         requests: histAsObj(requests, totalRequests),
         latency: addPercentiles(latencies, histAsObj(latencies)),
         throughput: histAsObj(throughput, totalBytes),
         errors: errors,
         timeouts: timeouts,
         duration: Math.round((Date.now() - startTime) / 1000),
+        start: startTime,
+        finish: Date.now(),
```
mcollina: I think we should have those as Date objects, rather than as epocs. Easier to print :).

thekemkid: Epochs are easier for transmission/serialisation though, right? I'll create them as objects though if you want.

mcollina: please do, those output will very likely be consumed by humans, and `Date` do jsonify.

thekemkid: done

mcollina: `start` is still an epoch, you will need to convert it to a `Date` object: https://github.com/mcollina/autocannon/blob/master/lib/run.js#L79

thekemkid: startTime needs to remain an epoch for the duration calculation, but I use new Date(startTime) to convert back to an object from the epoch. Do you want me to convert startTime to an object and then convert it to an epoch for the duration calculation? it's just two ways of achieving the same thing.

lib/run.js
```diff
@@ -139,12 +139,15 @@ function run (opts, cb) {
...
       let result = {
+        title: opts.title,
         requests: histAsObj(requests, totalRequests),
         latency: addPercentiles(latencies, histAsObj(latencies)),
         throughput: histAsObj(throughput, totalBytes),
         errors: errors,
         timeouts: timeouts,
         duration: Math.round((Date.now() - startTime) / 1000),
+        start: new Date(startTime),
+        finish: new Date(),
```
thekemkid: @mcollina Here. I'm using `new Date(value)`: https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Date

mcollina: oh I didn't see that. All good for me!

thekemkid: No problem, we were still talking on the outdated diff 👍

PR #45, opened 2016-07-06
pass headers and body of replies through client events

thekemkid:
for debugging purposes.


davidmarkclements:
I think in this case just "headers" and "body" events will do

thekemkid:
Cool, will update.

davidmarkclements:
also can add a "Debugging" header to the readme

that outlines this, and also listening for err objects

mcollina:
unit tests folks.

Issue #46, opened 2016-07-06
non 200's not recorded


davidmarkclements:
cc @thekemkid

mcollina:
fixed in #47

PR #47, opened 2016-07-06
now results have status codes

thekemkid:
Should this be considered the solution to #46 (non 200's not recorded)?

### DIFF
--- a/lib/run.js
+++ b/lib/run.js
@@ -99,8 +99,10 @@ function run (opts, cb) {
 
   function onResponse (statusCode, resBytes, responseTime) {
     tracker.emit('response', this, statusCode, resBytes, responseTime)
-    statusCodes[(parseInt(statusCode) / 100) - 1] += 1
-    latencies.record(responseTime)
+    const codeIndex = (parseInt(statusCode) / 100) - 1
+    statusCodes[codeIndex] += 1
+    // only record 2xx latencies
+    if (codeIndex === 1) latencies.record(responseTime)
     bytes += resBytes
     counter++
   }
--- a/lib/run.js
+++ b/lib/run.js
@@ -150,9 +150,9 @@ function run (opts, cb) {
         finish: new Date(),
         connections: opts.connections,
         pipelining: opts.pipelining,
-        '2xx': statusCodes[1],
         'non2xx': statusCodes[0] + statusCodes[2] + statusCodes[3] + statusCodes[4]
       }
+      statusCodes.forEach((code, index) => { result[(index + 1) + 'xx'] = code })
       tracker.emit('done', result)
 
       cb(null, result)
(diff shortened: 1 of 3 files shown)


davidmarkclements:
Yes - but to verify we need tests for 1xx, 3xx, 4xx, 5xx responses

test/run.test.js
```diff
@@ -179,3 +190,22 @@ test('run should callback with after a bailout', (t) => {
     t.end()
   }, 3000)
 })
+
+for (var i = 1; i <= 5; i++) closure(i)
```
davidmarkclements: we're node 4 and up, you can just use `let` instead of using func closure

test/helper.js
```diff
@@ -50,6 +50,25 @@ function startHttpsServer () {
   return server
 }
 
+for (var i = 1; i <= 5; i++) closure(i)
+
```
davidmarkclements: use `let`

test/run.test.js
```diff
@@ -179,3 +190,22 @@ test('run should callback with after a bailout', (t) => {
...
+
+function closure (i) {
+  test('run should count all reply status codes', (t) => {
```
davidmarkclements: change name of test to reflect error code, i.e. 

 ` test(`run should count all ${i}xx status codes`, (t) => {`

lib/run.js
```diff
@@ -150,9 +150,9 @@ function run (opts, cb) {
         finish: new Date(),
         connections: opts.connections,
         pipelining: opts.pipelining,
-        '2xx': statusCodes[1],
         'non2xx': statusCodes[0] + statusCodes[2] + statusCodes[3] + statusCodes[4]
```
davidmarkclements: I'm not sure we should keep non2xx

davidmarkclements:
couple of picks, but fine overall

I think while we're at it @thekemkid can we sort out #47 as it's semi-related

only 2xx should affect req per sec average and latency measurements

lib/run.js
thekemkid: It's used in tests, and its not very harmful to keep it in, I can remove if you prefer though

thekemkid:
uhm, #47 is very related as it's this PR... 😝 

Should I modify the code to only calc latency for 2xx's?

lib/run.js
mcollina: it's used in the cli output (non-json)

mcollina:
> Should I modify the code to only calc latency for 2xx's?

👍 Yes, please :)

mcollina:
LGTM

lib/run.js
davidmarkclements: Should  cli output all status codes though?

mcollina: Probably yes, the ones that it finds

PR #49, opened 2016-07-06
kOnHeaders bugfix

davidmarkclements:
If a response has trailers, it will trigger the following:

https://github.com/creationix/http-parser-js/blob/master/http-parser.js#L353

We need a noop for kOnHeaders otherwise autocannon will throw when a response has trailers

cc @mcollina


mcollina:
Unit test? :)

davidmarkclements:
added :bowtie:

mcollina:
LGTM

PR #51, opened 2016-07-06
can specify max reqs (both overall and per connection)

### LINKED ISSUE #50: requests per connection option
overrides duration, set amount of requests to make per conn, and finish when all requests are complete

for instance, "I want to see how long it take for 100 connection to concurrently make one request"

thekemkid:
Fixes #50


lib/run.js
```diff
@@ -87,6 +99,13 @@ function run (opts, cb) {
   url.setupClient = opts.setupClient
   url.timeout = opts.timeout
   url.requests = opts.requests
+  url.responseMax = opts.maxConnectionRequests
+
+  if (!url.responseMax) {
+    url.responseMax = typeof opts.maxOverallRequests === 'number'
+                      ? (opts.maxOverallRequests / opts.connections) : false
+    // this needs work if ((maxOverallRequests % opts.connections) > 0)
+  }
```
thekemkid: I expect you'll see this. If the response max is set with the MaxConnectionReqs this will be jumped over, as MaxConnectionReqs takes precedence. If it is not set, we check if the maxOverallReqs is a number to avoid a NAN division error. I originally had:
```
url.responseMax = opts.maxConnectionRequests
                      || typeof opts.maxOverallRequests === 'number'
                          ? (opts.maxOverallRequests / opts.connections)
                          : false
```
but encountered an error were url.responseMax was set to 0, when opts.maxConnectionRequests was passed in.

mcollina: I don't like that comment over `maxOverallRequests % opts.connections > 0`.  Sort it out, or `cb(new Error('maxOverallRequest must be multiple of the number of connections')`.

lib/httpClient.js
```diff
@@ -105,6 +107,10 @@ Client.prototype._connect = function () {
 
 // rpi = request pipelining index
 Client.prototype._doRequest = function (rpi) {
+  if (!this.destroyed && this.responseMax && this.reqsMade++ >= this.responseMax) {
+    this.destroy()
+    return
+  }
```
mcollina: Does this affects performance anyhow?

lib/run.js
mcollina: You should add a comment on all of the above in the code, your sentence explain it, but otherwise it is not clear. I think that setting `responseMax` to `false` is generally a bad idea, because we say it is a number in the README.

lib/httpClient.js
thekemkid: nope perf is the same in my benchmark runs

lib/run.js
thekemkid: The throw seems like a smarter idea, will do. I'll set it to 0 then, as that's falsey.

davidmarkclements:
so this has to override duration - duration has to be determined by all requests completing

davidmarkclements:
I don't think we need per connection, all we need is to implement this in run.js, and instead of stopping after a certain amount of time, stop after a certain amount of requests

thekemkid:
@mcollina && @davidmarkclements I think that this is finished and ready for merge and publish, but it needs review first

mcollina:
LGTM for me, but let's wait for @davidmarkclements

davidmarkclements:
I don't think we're quite ready for merge - this affects the progress bar output, 
it says "Running 10s test" - where it should "Running 10reqs test" or whatever,
and the progress bar is based on 10s (default duration) instead of amount of
completed requests

thekemkid:
Nice catch. Done.

mcollina:
LGTM

PR #57, opened 2016-07-11
graceful redirection of output

thekemkid:


lib/progressTracker.js
```diff
@@ -23,6 +23,9 @@ function track (instance, opts) {
 
   opts = xtend(defaults, opts)
 
+  // don't render tracking info if stdout is a tty and render json is on
+  if (opts.json && process.stdout.isTTY) return
```
mcollina: can you move this back into the main file?

mcollina:
LGTM, merge and release (patch level).

PR #61, opened 2016-07-12
Avoid printing the progress bar if the destination is not a TTY.

### LINKED ISSUE #60: error when redirecting output
Under Linux, with node 6.3.0. Also happens on master:

`autocannon -d 1 example.com > out 2>&1`

Result - error in node_modules/progress/lib/node-progress.js:177: 

`TypeError: this.stream.clearLine is not a function`

Happens only if both stdout and stderr are redirected.

mcollina:
Fixes #60.

@davidmarkclements @thekemkid please review.

### DIFF
--- a/lib/progressTracker.js
+++ b/lib/progressTracker.js
@@ -26,54 +26,60 @@ function track (instance, opts) {
   const iOpts = instance.opts
   if (opts.renderProgressBar) {
     let msg = `${iOpts.connections} connections`
+
     if (iOpts.pipelining > 1) {
       msg += ` with ${iOpts.pipelining} pipelining factor`
     }
-    if (!iOpts.amount) {
-      const progressBar = new ProgressBar(opts.progressBarString, {
-        width: 20,
-        incomplete: ' ',
-        total: iOpts.duration,
-        clear: true,
-        stream: opts.outputStream
-      })
 
+    if (!iOpts.amount) {
       logToStream(`Running ${iOpts.duration}s test @ ${iOpts.url}\n${msg}\n`)
 
-      progressBar.tick(0)
+      // if switch needed needed to avoid
+      // https://github.com/mcollina/autocannon/issues/60
+      if (opts.outputStream.isTTY) {
+        const progressBar = new ProgressBar(opts.progressBarString, {
+          width: 20,
+          incomplete: ' ',
+          total: iOpts.duration,
+          clear: true,
+          stream: opts.outputStream
+        })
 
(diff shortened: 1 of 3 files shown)


lib/progressTracker.js
```diff
@@ -26,54 +26,60 @@ function track (instance, opts) {
...
-      instance.on('tick', () => {
-        progressBar.tick()
-      })
-
-      process.once('SIGINT', () => {
-        progressBar.tick(iOpts.duration - 1)
-      })
+      // if switch needed needed to avoid
+      // https://github.com/mcollina/autocannon/issues/60
+      if (opts.outputStream.isTTY) {
```
thekemkid: We could turn the progress bar output into its own function, and in that case, we could return early if the output stream isn't a tty, like below

```js
if (!opts.outputStream.isTTY) return
```

Not a worry if you don't want to, only a minor thing I realised

test/cli.test.js
```diff
@@ -0,0 +1,44 @@
...
+const lines = [
+  /Running 1s test @ .*$/,
+  /10 connections.*$/,
+  /$/,
+  /Stat.*Avg.*Stdev.*Max.*$/,
+  /Latency \(ms\).*$/,
+  /Req\/Sec.*$/,
+  /Bytes\/Sec.*$/,
+  /$/,
+  /\d+k requests in 1s, .* MB read$/
```
thekemkid: I know it _should_ be 1s, it _could_ be longer than 1s... could we change it to a selection group of [1|2]? any longer is a fail for sure :)

test/cli.test.js
```diff
@@ -0,0 +1,44 @@
...
+const lines = [
+  /Running 1s test @ .*$/,
+  /10 connections.*$/,
+  /$/,
+  /Stat.*Avg.*Stdev.*Max.*$/,
+  /Latency \(ms\).*$/,
+  /Req\/Sec.*$/,
+  /Bytes\/Sec.*$/,
+  /$/,
+  /.* requests in 1s, .* read/
```
thekemkid: I know that it _should_ be 1s, it _could_ be longer than 1s... could we change it to a selection group of [1|2]? any longer is a fail for sure :)

lib/progressTracker.js
mcollina: what we should probably do is to wrap those two blocks in functions, the main track function is hardly readable.

thekemkid: Right, you can merge and I'll do that later :)

PR #69, opened 2016-07-14
prepend url with http:// by default if it isn't there, and output url  in result json

### LINKED ISSUE #68: Should we ensure strings start with `http`?
I ask because if you do something like `autocannon google.com` it doesn't recognise the actual address. Simple enough to fix, just check `url.indexOf('http') === 0` and if not, prepend 'http://' before it's parsed by node's internal url module.

This check would also pass for `https`, hence why I don't check for `http://`

thekemkid:
Fixes #68

### DIFF
--- a/lib/run.js
+++ b/lib/run.js
@@ -49,6 +49,7 @@ function run (opts, cb) {
   // is done
   tracker.opts = opts
 
+  if (opts.url.indexOf('http') !== 0) opts.url = 'http://' + opts.url
   const url = URL.parse(opts.url)
 
   let counter = 0
@@ -153,6 +154,7 @@ function run (opts, cb) {
       clients.forEach((client) => client.destroy())
       let result = {
         title: opts.title,
+        url: opts.url,
         requests: histAsObj(requests, totalRequests),
         latency: addPercentiles(latencies, histAsObj(latencies)),
         throughput: histAsObj(throughput, totalBytes),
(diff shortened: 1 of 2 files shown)


lib/run.js
```diff
@@ -49,6 +49,7 @@ function run (opts, cb) {
   // is done
   tracker.opts = opts
 
+  if (opts.url.indexOf('http') !== 0) opts.url = 'http://' + opts.url
```
mcollina: probably we should error nicely if url is missing.

test/run.test.js
```diff
@@ -204,6 +204,19 @@ test('run should allow users to enter timestrings to be used for duration', (t)
...
+
+  run({
+    url: 'localhost:' + server.address().port,
+    duration: 1
+  }, (err, res) => {
+    t.error(err)
+    t.equal(res.url, 'http://localhost:' + server.address().port, 'url should have http:// added to start')
+    t.end()
+  })
+})
```
thekemkid: If there are no results, this will fail, as res.url will throw `cant access 'url' on undefined`, or something of the like. Will also add a t.ok(res) though :)

lib/run.js
thekemkid: We do, check line 179

thekemkid: https://github.com/thekemkid/autocannon/blob/00310881b982e47950917cacb92d719be81d7524/lib/run.js#L179

PR #75, opened 2016-07-22
added reconnect ability and documented it

### LINKED ISSUE #65: disconnect and reconnect clients after a while
We should be able to force a dns lookup each X requests to support AWS ELBs:

> Single Client Tests
In single client tests, a load generator will issue requests to the server using the same client. In most cases, the client does not re-resolve the DNS after the test is initiated. A good example of this type of test is Apache Bench (ab). If you are using this type of testing tool, there are two options:
> 
> * Write discrete tests at a given sample size and concurrency then start a new set of tests and invoke these tests separately so that the client will re-resolve the DNS (possibly even forcing a re-resolution depending on the operating system and other factors). It is important to save detailed request logs so you can aggregate the logs for a single view of the request details.
> * Launch multiple client instances and initiate the tests on the different instances at approximately the same time. AWS CloudFormation can help you do this by creating your load test script and launching multiple clients, then simply issuing remote SSH commands to initiate the tests. You may want to bring up more load generating clients based on Auto Scaling rules (for example, when the average CPU of other load clients reaches a certain level, add more load generators).
>
> However, depending on the load you are trying to generate, your test may be constrained by the client's ability to issue requests. In this case, you need to consider distributed tests.

thekemkid:
Fixes #65


autocannon.js
```diff
@@ -32,6 +32,7 @@ function start () {
       maxOverallRequests: 'O',
       connectionRate: 'r',
       overallRate: 'R',
+      resetRate: 'D',
```
mcollina: I wouldn't call "resetRate", but rather "reconnectAfter".

README.md
```diff
@@ -81,6 +81,9 @@ Available options:
         connection rate will take precedence if both are set.
         NOTE: if using rate limiting and a very large rate is entered which cannot be met,
               Autocannon will do as many requests as possible per second.
+  -D/--resetRate NUM
+        Some number of requests to make before resetting a connections connection to the
+        server.
```
mcollina: I wouldn't call it `reset`, it's rather a reconnect. `reconnectRate` is ok.

lib/httpClient.js
```diff
@@ -121,22 +125,28 @@ Client.prototype._connect = function () {
...
     }
     this.emit('request')
-
     this.resData[rpi].startTime = process.hrtime()
     this.conn.write(this.requestIterator.move())
     this.timeoutTicker.reschedule(this.timeout)
+    this.reqsMade++
+    if (!this.destroyed && this.resetRate && this.reqsMade % this.resetRate === 0) {
+      return this._resetConnection()
+    }
```
mcollina: I don't think you should reconnect here, but rather when you get a reply.

thekemkid: Ok, will move 👍

mcollina:
LGTM once travis agrees :P.

PR #76, opened 2016-07-22
print out ndjson instead of prettyfied json

thekemkid:


mcollina:
LGTM and bump the major.

PR #80, opened 2016-08-02
better color support testing

### LINKED ISSUE #59: colors disabled when -j is set and we are writing to a file
As titled, we might probably want to override some stuff in [`chalk`](https://github.com/chalk/chalk).

Chalk is disabled because of a check in https://github.com/chalk/supports-color/blob/master/index.js.

See https://github.com/chalk/supports-color/blob/master/index.js#L41-L43.

We can either internalize that check, or sending them a PR to expose that as a function that accepts a stream. Or maybe it's just a new module.

thekemkid:
Fixes #59

I added color-support as a dependency as it accomplishes what we need.

### DIFF
--- a/lib/progressTracker.js
+++ b/lib/progressTracker.js
@@ -2,7 +2,8 @@
 
 const ProgressBar = require('progress')
 const table = require('table')
-const chalk = require('chalk')
+const Chalk = require('chalk')
+const testColorSupport = require('color-support')
 const prettyBytes = require('pretty-bytes')
 const xtend = require('xtend')
 const format = require('./format')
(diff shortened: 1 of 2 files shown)


lib/progressTracker.js
```diff
@@ -156,7 +161,7 @@ function asRow (name, stat) {
   ]
 }
 
-function asColor (color, row) {
+function asColor (chalk, color, row) {
```
mcollina: A nicer way to do it is to pass chalk.color directly as a function. It would probably be nicer.

mcollina:
Nice one! Just one nit!

mcollina:
LGTM

Issue #82, opened 2016-08-08
TypeError: Expected a number, got number


bzoz:
When benchmarking [this node.js script](https://gist.github.com/bzoz/4bedbd51c3160875144c1dd108673eaa) with:
`autocannon -d 1 localhost:1234`

Result:
```
Running 1s test @ http://localhost:1234
10 connections

/usr/local/lib/node_modules/autocannon/node_modules/pretty-bytes/index.js:6
		throw new TypeError('Expected a number, got ' + typeof num);
		^

TypeError: Expected a number, got number
    at module.exports (/usr/local/lib/node_modules/autocannon/node_modules/pretty-bytes/index.js:6:9)
    at asBytes (/usr/local/lib/node_modules/autocannon/lib/progressTracker.js:170:20)
    at EventEmitter.instance.on (/usr/local/lib/node_modules/autocannon/lib/progressTracker.js:59:38)
    at emitOne (events.js:101:20)
    at EventEmitter.emit (events.js:188:7)
    at Timeout.setInterval (/usr/local/lib/node_modules/autocannon/lib/run.js:181:15)
    at Timeout.wrapper (timers.js:425:11)
    at tryOnTimeout (timers.js:232:11)
    at Timer.listOnTimeout (timers.js:202:5)
```
This does not happen, when `-j` is used, which produces:
```
{"url":"http://localhost:1234","requests":{"average":1680,"mean":1680,"stddev":0,"min":1680,"max":1680,"total":1680,"sent":1690},"latency":{"average":5.54,"mean":5.54,"stddev":2.36,"min":3,"max":19,"p50":4,"p75":7,"p90":9,"p99":14,"p999":17,"p9999":19,"p99999":19},"throughput":{"average":null,"mean":null,"stddev":null,"min":-1,"max":0,"total":1761814320},"errors":0,"timeouts":0,"duration":1,"start":"2016-08-08T10:58:21.518Z","finish":"2016-08-08T10:58:22.540Z","connections":10,"pipelining":1,"non2xx":0,"1xx":0,"2xx":1680,"3xx":0,"4xx":0,"5xx":0}
```

thekemkid:
Unfortunately I'm having some trouble reproducing this, can you confirm this is happening with current master?

mcollina:
Maybe this is related: https://github.com/mcollina/autocannon/pull/83

bzoz:
@thekemkid Tried with master, same results. 

[...]

PR #86, opened 2016-08-16
Bump max throughput to 100GB/s

bzoz:
Bumps max throughput to 100GB/s.

Fixes: https://github.com/mcollina/autocannon/issues/82


mcollina:
Thanks!!

PR #94, opened 2016-11-08
Add functionality to run autocannon forever.

### LINKED ISSUE #92: feature_request: make autocannon run indefinitely whilst emitting reports periodically
In a microservice architecture it is nice to have constant load on the systems to track p99 latencies.

This talk gives a nice real world example done by Uber. https://www.youtube.com/watch?v=kb-m2fasdDY

Hence it would be cool to have:

``` js
const instance = autocannon({
  url: 'https://url.url',
  connections: 30, 
  pipelining: 3,
  duration: 0 // for safe infinite
}, console.log)

instance.on('report', sendToAnalytics)
```

thekemkid:
+ tests & documentation

fixes #92 

No perf damage over multiple runs.

### DIFF
--- a/lib/progressTracker.js
+++ b/lib/progressTracker.js
@@ -55,14 +55,14 @@ function track (instance, opts) {
   // added on restarting, causing listener leaks
   if (opts.renderProgressBar && opts.outputStream.isTTY) {
     if (!iOpts.amount) { // duration progress bar
-      instance.on('tick', curriedTick(durationProgressBar))
-      instance.on('done', curriedTick(durationProgressBar, iOpts.duration - 1))
-      process.once('SIGINT', curriedTick(durationProgressBar, iOpts.duration - 1))
+      instance.on('tick', () => { durationProgressBar.tick() })
+      instance.on('done', () => { durationProgressBar.tick(iOpts.duration - 1) })
+      process.once('SIGINT', () => { durationProgressBar.tick(iOpts.duration - 1) })
     } else { // amount progress bar
-      instance.on('response', curriedTick(amountProgressBar))
-      instance.on('reqError', curriedTick(amountProgressBar))
-      instance.on('done', curriedTick(amountProgressBar, iOpts.amount - 1))
-      process.once('SIGINT', curriedTick(amountProgressBar, iOpts.amount - 1))
+      instance.on('response', () => { amountProgressBar.tick() })
+      instance.on('reqError', () => { amountProgressBar.tick() })
+      instance.on('done', () => { amountProgressBar.tick(iOpts.amount - 1) })
+      process.once('SIGINT', () => { amountProgressBar.tick(iOpts.amount - 1) })
     }
   }
 
@@ -176,10 +176,6 @@ function asBytes (stat) {
   result.max = prettyBytes(stat.max)
   result.min = prettyBytes(stat.min)
   return result
-}
-
-function curriedTick (progressBar, amount) {
-  return () => { progressBar.tick(amount) }
 }
 
 module.exports = track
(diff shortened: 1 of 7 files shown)


thekemkid:
Weird, I actually fixed the test issue locally on my machine before pushing, but it is appearing again on CI. investigating

thekemkid:
*sigh*. Nothing happened here. Move along.

eljefedelrodeodeljefe:
Awesome

test/forever.test.js
```diff
@@ -0,0 +1,28 @@
...
+    t.equal(res.duration, 1, 'should have take 1 seconds to run')
+    numRuns++
+    if (numRuns === 2) {
+      instance.stop()
+      setTimeout(() => {
+        t.ok(true, 'should have reached here without the callback being called again')
+        t.end()
+      }, 1000)
+    }
+  })
```
mcollina: A callback should be called only once, we should probably move away from the callback, and use an EventEmitter.

lib/progressTracker.js
```diff
@@ -24,26 +24,45 @@ function track (instance, opts) {
...
+  if (opts.renderProgressBar && opts.outputStream.isTTY) {
+    if (!iOpts.amount) { // duration progress bar
+      instance.on('tick', () => { durationProgressBar.tick() })
+      instance.on('done', () => { durationProgressBar.tick(iOpts.duration - 1) })
+      process.once('SIGINT', () => { durationProgressBar.tick(iOpts.duration - 1) })
+    } else { // amount progress bar
+      instance.on('response', () => { amountProgressBar.tick() })
+      instance.on('reqError', () => { amountProgressBar.tick() })
+      instance.on('done', () => { amountProgressBar.tick(iOpts.amount - 1) })
+      process.once('SIGINT', () => { amountProgressBar.tick(iOpts.amount - 1) })
```
mcollina: can we factor this out into a function?

thekemkid: We could, but it should only be run once, so I left it in the scope of this function to make it clear to readers that this is should only be run once, when the track function is ran.

test/forever.test.js
thekemkid: So, this is a weird case, and I only planned on using the event emitter approach. But I realised that this is possible when writing the test so I used it. In the case `forever` is true, should we throw if a callback is passed in?

lib/progressTracker.js
mcollina: I mean, the code above for both `amountProgressBar` and `durationProgressBar` looks _very_ similar. we should probably factor it out in its own function, ideally calling the same function with different parameters, or something similar.

(this is a nit of style, nothing more)

test/forever.test.js
mcollina: Yes, I think so.

lib/progressTracker.js
thekemkid: I could curry the callbacks for the events with the relevant progress bar var & amount to tick by (undefined in the case of raw `.tick()`). Thoughts?

test/forever.test.js
thekemkid: This adds much more complexity when handling options. Have it working, will push so you can review

lib/progressTracker.js
thekemkid: Okay, so I did the currying thing, and felt super happy with it, and then realised why I had them written like that when it crashed. I use the scope of the progressBar variables to instantiate new ones when restarting the benchmark, but use the same variable reference. By currying, I lose access to the scope-referencing-closure-benefit. I would like to leave this as is, because to get something "cleaner"  working would end up adding add much more code, and not necessarily add any benefit. Is that okay with you?

mcollina: Yeah, but put in a comment.

thekemkid: comment in.

thekemkid:
If you're happy with this @mcollina, feel free to merge. I don't think I can add anymore to it 👍

PR #95, opened 2016-11-16
Adding PATCH

redlightmikey:
I needed PATCH for testing a PATCH based API. 

Tested this manually with my PATCH API.


davidmarkclements:
nice thanks @redlightmikey !

thekemkid:
I guess all http verbs should be valid here. Will do!

PR #96, opened 2016-11-16
Added all http methods/verbs

thekemkid:
I listed them in the order they're presented here: https://en.wikipedia.org/wiki/Hypertext_Transfer_Protocol#Summary_table

### DIFF
--- a/lib/httpMethods.js
+++ b/lib/httpMethods.js
@@ -1,0 +1,13 @@
+// most http methods taken from RFC 7231
+// PATCH is defined in RFC 5789
+module.exports = [
+  'GET',
+  'HEAD',
+  'POST',
+  'PUT',
+  'DELETE',
+  'CONNECT',
+  'OPTIONS',
+  'TRACE',
+  'PATCH'
+]
--- a/lib/httpRequestBuilder.js
+++ b/lib/httpRequestBuilder.js
@@ -2,11 +2,17 @@
 
 const xtend = require('xtend')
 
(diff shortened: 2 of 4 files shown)


mcollina:
LGTM

mcollina:
Seems tests are failing

thekemkid:
For some reason we test to ensure 'client does not supports HEAD'

Why?

mcollina:
Maybe we don't support it. LOL.

thekemkid:
But... why wouldn't we support it? Why do we explicitly test for non-support? 😂

There's more lines of code in testing for non-support rather than _just_ supporting it. 😂

mcollina:
I _remember_ that I had *good reasons* not to do it, so important reasons that I completely forgot about them.

thekemkid:
😂

I have a good reason to support it: We are a http 1.1 bench marker. we should support it :P

The final call is up to you, should I fix the tests, or remove HEAD support? These are all which are supported by wrk: https://github.com/wg/wrk/blob/master/src/http_parser.c#L960-L974

mcollina:
please go ahead.

thekemkid:
I assume go ahead is to mean fix tests... on it! 😄

thekemkid:
@mcollina I know you have been so excited for this update... but its ready for a re-review! 😄 I moved the valid array of http methods to their own file, for testability purposes. Hope this is okay 👍

PR #106, opened 2017-04-14
Fix `--latency` option

drifkin:
The code for this option wasn't modified for breaking changes introduced
when deps were updated in ea237e343c0081bbf5725b96f8c3d8cea9b495cd


mcollina:
Thanks :)

PR #114, opened 2017-10-09
README: remove non existent benchmarks header from toc

anchnk:
This patch removes Benchmarks header link from table of content as the
section isn't present anymore.


mcollina:
THanks!!

PR #116, opened 2018-01-25
Removes the binary dependency

mcollina:
As titled, this switches from native-hdr-histogram to hdr-histogram-js.
In my tests, the throughput is about the same, if not slightly higher.
It solves all the problems downloading binary components.


.travis.yml
```diff
@@ -4,6 +4,8 @@ node_js:
   - "4"
   - "6"
   - "7"
+  - "8"
+  - "9"
 env:
   - CC=gcc-4.8 CXX=g++-4.8
```
thekemkid: We should actually be able to remove these build tools, right?

mcollina:
This should be ready to land.

PR #117, opened 2018-01-29
Fix bytes/sec max NaN output

watson:
`stat.max` have already been converted to a human readable string containing the appropriate unit, so there's no need to format it further.

### DIFF
--- a/lib/progressTracker.js
+++ b/lib/progressTracker.js
@@ -164,7 +164,7 @@ function asRow (name, stat) {
     name,
     stat.average,
     stat.stddev,
-    Math.floor(stat.max * 100) / 100
+    stat.max
   ]
 }
 


mcollina:
my bad

mcollina:
However I get this:

```
$ ./autocannon.js -c 100 -d 5 -p 10 localhost:3000
Running 5s test @ http://localhost:3000
100 connections with 10 pipelining factor

Stat         Avg     Stdev   Max
Latency (ms) 4.18    13.6    288.021199
Req/Sec      23457.6 3444.77 25824
Bytes/Sec    2.6 MB  370 kB  2.87 MB

117k requests in 5s, 13 MB read
```

Can you send another PR that prevents that?

watson:
Ah, didn't see that. Hang on

mcollina:
I would like it rounded at 2 digits.

watson:
Makes sense

watson:
@mcollina fixed in #118

PR #119, opened 2018-01-29
Use autocannon on a UNIX socket

watson:
With this you can tell autocannon to send requests to a UNIX socket instead of a TCP socket, e.g:

```
autocannon /tmp/my-server.sock
```

For use with test HTTP server:

```js
const http = require('http')
http.createServer(function (req, res) {
  res.end('Hello World')
}).listen('/tmp/my-server.sock')
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ npm i autocannon --save
 Usage: autocannon [opts] URL
 
 URL is any valid http or https url. Can alternatively be a path to a
-UNIX socket.
+Unix Domain socket.
 
 Available options:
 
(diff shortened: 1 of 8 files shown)


watson:
@mcollina before I dive deeper into this rabbit hole, I just wanted to hear your opinion on this - if this is even something you'd like to add to autocannon.

My use-case is that I'm benchmarking the difference between letting an HTTP server listen on a TCP socket vs a UNIX socket to see how much more throughput I can get this way for use with interprocess communication.

watson:
@mcollina updated

PR #121, opened 2018-01-30
Clean up use of teardown

watson:
The new tap v11.1.0 improves the ability to disable `autoend`, which means we can use t.tearDown even if we don't know call t.plan again.

See:
https://github.com/tapjs/node-tap/commit/fcf70aad8582d434658b8d3fe794e6ee463bcd2d

P.S. this was JUST released, so I unfortunately couldn't have included this in my last PR 😅


mcollina:
Thanks!

PR #123, opened 2018-02-21
Allow use of http host and path during IPC

watson:
Previously the Host header was always `localhost` when connecting to a Unix Domain Socket or a Windows Named Pipe. It was also not possible to use any other path than `/`.

With this commit the URL is always treated as a URL and if you want to supply a Unix Domain Socket or Windows Named Pipe, you instead have to provide it using `--socketPath` (aliased `-S`).

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,9 +40,6 @@ Usage: autocannon [opts] URL
 
 URL is any valid http or https url.
 
-For IPC, the url can be substituted by a path to a Unix Domain Socket or
-a Windows Named Pipe.
-
 Available options:
 
   -c/--connections NUM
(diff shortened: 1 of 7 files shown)


watson:
In its current state this PR should be considered a breaking change. I can make it so that it falls back to the old behavior if the URL provided looks like a socket/pipe if we don't want this to bump the major.

watson:
@mcollina The tests are failing, but it seems to be an issue with Travis rather with my perfect code 😜 . I don't have permissions to restart the test though

thekemkid:
@watson I've restarted it for you :)

thekemkid:
I'd like to leave the decision to @mcollina what he wants to do with the PR. I'm inclined to just merge and bump major. 👍

PR #125, opened 2018-03-02
fix default option issue

WLBF:
I get something wrong when I run two autocannon benchmark at same time. However if I run these benchmarks separately it works fine.
example:
```js
const autocannon = require('autocannon');
const util = require('util');

const autocannonFunc = util.promisify(autocannon);

function setupClient {
    ...
}

async function login() {
  try {
    return await autocannonFunc({
      url: 'http://192.168.0.245:10000/login',
      connections: 15,
      pipelining: 1,
      amount: 15,
      bailout: 1,
      method: 'POST',
      setupClient,
    });
  } catch(err) {
      ...
  }
}

async function register() {
  try {
    return await autocannonFunc({
      url: 'http://192.168.0.245:8888/register',
      connections: 1,
      pipelining: 10,
      amount: 20,
      bailout: 1,
      method: 'POST',
      headers: {
        'Content-Type': 'application/json',
      },
      body: JSON.stringify(body),
    });
  } catch(err) {
      ...
  }
}

Promise.all([login(), register()])
```
run above code ,  second autocannon instance will return abnormal 5xx code . I figure out that first calling of  `run()` function change the `requests` field of `defaultOptions` object, cause second calling failed. So I think it should be fine to move `defaultOptions` into function `run()` block to avoid the side effect.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,6 +44,7 @@
   },
   "dependencies": {
     "chalk": "^2.0.0",
+    "clone": "^2.1.1",
     "color-support": "^1.1.1",
     "deep-extend": "^0.5.0",
     "hdr-histogram-js": "^1.0.0",
--- a/test/run.test.js
+++ b/test/run.test.js
@@ -3,7 +3,9 @@
 const os = require('os')
 const path = require('path')
 const test = require('tap').test
+const clone = require('clone')
 const run = require('../lib/run')
+const defaultOptions = require('../lib/defaultOptions')
 const helper = require('./helper')
 const server = helper.startServer()
 
@@ -279,3 +281,16 @@ for (let i = 1; i <= 5; i++) {
     })
   })
 }
+
+test('run should not modify default options', (t) => {
+  const origin = clone(defaultOptions)
+  run({
+    url: 'http://localhost:' + server.address().port,
+    connections: 2,
+    duration: 2
+  }, function (err, result) {
+    t.error(err)
+    t.deepEqual(defaultOptions, origin, 'calling run function does not modify default options')
+    t.end()
+  })
+})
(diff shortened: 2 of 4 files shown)


mcollina:
I think the proper fix is to switch from `xtend` to `Object.assign` in https://github.com/mcollina/autocannon/pull/125/files#diff-de78b50723ff596d7bd35665f36a6cbfR73.

Can you also add a unit test for this?

WLBF:
sure, I am working on it.

WLBF:
I think `Object.assign` only copy first layer of a object. Here we need do full deep copy of `defaultOptions` object to  avoid modification come from `run()` function. I also adjust the position of `defaultOptions` to add a unit test.

package.json
```diff
@@ -44,7 +44,9 @@
   },
   "dependencies": {
     "chalk": "^2.0.0",
+    "clone": "^2.1.1",
     "color-support": "^1.1.1",
+    "deep-equal": "^1.0.1",
```
mcollina: there is no need for this dependency, tap can do it for us.

test/run.test.js
```diff
@@ -279,3 +282,16 @@ for (let i = 1; i <= 5; i++) {
...
+
+test('run should not modify default options', (t) => {
+  const origin = clone(defaultOptions)
+  run({
+    url: 'http://localhost:' + server.address().port,
+    connections: 2,
+    duration: 2
+  }, function (err, result) {
+    t.error(err)
+    t.ok(deepEqual(defaultOptions, origin), 'calling run function does not modify default options')
```
mcollina: you can use t.deepEqual here.

WLBF:
ok

PR #128, opened 2018-04-18
Added optional server name for the SNI TLS extension

sebbarg:
Server Name Indication (SNI) is an extension to TLS by which a client indicates which hostname it is attempting to connect to at the start of the handshaking process. This allows a server to present multiple certificates on the same IP address.

https://en.wikipedia.org/wiki/Server_Name_Indication

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,6 +90,8 @@ Available options:
         Print the output as newline delimited json. This will cause the progress bar and results not to be rendered. default: false.
   -f/--forever
         Run the benchmark forever. Efficiently restarts the benchmark on completion. default: false.
+  -s/--sniServerName
+        Server name for the SNI (Server Name Indication) TLS extension.
   -v/--version
         Print the version number.
   -h/--help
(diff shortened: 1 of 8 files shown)


mcollina:
Can you please add a unit test?

sebbarg:
Hi Matteo,

The only thing it does is pass along the server name to the Node
tls.connect() function.

t's not trivial to test that it actually works in a unit test since it
would require setting up a much more advanced web server / reverse proxy
with multiple certificates that would know what to do with the SNI
handshake.

Or maybe you have something else in mind?

mcollina:
It's not just a matter of specifying a `SNICallback` when a server is created https://nodejs.org/api/tls.html#tls_tls_createserver_options_secureconnectionlistener?

README.md
```diff
@@ -90,6 +90,8 @@ Available options:
         Print the output as newline delimited json. This will cause the progress bar and results not to be rendered. default: false.
   -f/--forever
         Run the benchmark forever. Efficiently restarts the benchmark on completion. default: false.
+  -s/--sniServerName
```
mcollina: I would prefer if this was `-S/--servername`.

sebbarg:
Sure thing - capital S is already in use, so it has to be lowercase.

sebbarg:
I will have a look

sebbarg:
All done. The unit test tests whether the server name is sent correctly
through to tje https / tls connection. Since it is working on tcp level
rather than http, the test server constructs a http response with the
server name in the header. In that way the client side of the unit test can
reuse the http functionality to verify if the name was correctly sent
through.

PR #129, opened 2018-04-27
Support excluding error statistics from latency and bytes per second averages

### LINKED ISSUE #127: latency is NaN for non-2xx responses
```sh
$ node -e "http.createServer((req, res) => { res.statusCode=400; res.end()}).listen(3000)"
```

```sh
$ autocannon localhost:3000
Running 10s test @ http://localhost:3000
10 connections

Stat         Avg     Stdev     Max
Latency (ms) NaN     NaN       0
Req/Sec      21121.6 898.21    22575
Bytes/Sec    2.28 MB 101.58 kB 2.49 MB

0 2xx responses, 211216 non 2xx responses
211k requests in 10s, 22.81 MB read
```

thekemkid:
Fixes #127

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,6 +84,8 @@ Available options:
         Don't render the progress bar. default: false.
   -l/--latency
         Print all the latency data. default: false.
+  -I/--idReplacement
+        Enable replacement of [<id>] with a randomly generated ID within the request body. default: false.
   -j/--json
         Print the output as newline delimited json. This will cause the progress bar and results not to be rendered. default: false.
   -f/--forever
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -84,14 +84,14 @@ Available options:
         Don't render the progress bar. default: false.
   -l/--latency
         Print all the latency data. default: false.
-  -I/--idReplacement
-        Enable replacement of [<id>] with a randomly generated ID within the request body. default: false.
```
mcollina: why are you removing this?

thekemkid: It turns out this was actually missing from the `help.txt` file, and I copied the contents of that over during this PR. This is missing there too. I'll add it to that file, and fix this 👍

PR #131, opened 2018-05-02
Pass error to reqError listener

### LINKED ISSUE #111: reqError callback throws nothing
```js

  function finishedBench() {
    console.log('finished benchmarking');
    // console.log(res);
    // console.log('finished bench', err, res);
  }

  const config = {
    url: 'http://localhost:3001/5968537fe66a6431eaa81ff8/5987c22fcea4da29f1f97b29',
    amount: 54000,
    timeout: 5400,
    connections: 3000,
  };
  const instance = autocannon(config, finishedBench);

  instance.on('reqError', (...args) => {
    console.log('---------------------');
    console.log(args);
    console.log('---------------------');
  });
```
Managed to trigger some errors.

but this is what i seen on terminal

```sh
...
[]
---------------------
---------------------
[]
---------------------
---------------------
[]
---------------------
---------------------
[]
...
```
it seems in trigger the **reqError** event, but it misses the "details" of the error.

but when i adjust the config to 
```js
  const config = {
    url: 'http://localhost:3001/5968537fe66a6431eaa81ff8/5987c22fcea4da29f1f97b29',
    amount: 54000,
    timeout: 5400,
    connections: 2000,
  };
```
there is no **reqError** event emitted. :+1: 

Questions

1. is the 3k concurrent connections quite suffocating to my app?
2. why the reqError event is throwing empty?
3. on connections,
` 
 -c/--connections NUM
        The number of concurrent connections to use. default: 10.
`, does this mean that the autocannon will try its very best to create the connections and fire them simultenously to my app?
4. im developing an app that will handle 8M requests  and generate 8M records at least per day, what will be the ideal configurations/settings i must use against it?

doesdev:
This fixes #111

### DIFF
--- a/lib/run.js
+++ b/lib/run.js
@@ -217,15 +217,16 @@ function run (opts, cb) {
       counter++
     }
 
-    function onError () {
-      for (let i = 0; i < opts.pipelining; i++) tracker.emit('reqError')
+    function onError (error) {
+      for (let i = 0; i < opts.pipelining; i++) tracker.emit('reqError', error)
       errors++
       if (opts.bailout && errors >= opts.bailout) stop = true
     }
 
     // treat a timeout as a special type of error
     function onTimeout () {
-      for (let i = 0; i < opts.pipelining; i++) tracker.emit('reqError')
+      let error = new Error('request timed out')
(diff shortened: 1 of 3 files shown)


mcollina:
Thanks!!!

PR #134, opened 2018-05-14
Support standard syntax for HTTP headers

### LINKED ISSUE #133: autocannon should support standard definitions of http headers
hi,

i was using `autocannon` to stress test my web application and i find out that `autocannon` requires passing custom HTTP headers using a non standard format, specifically `-H "Content-Type=application/json"`, that is not conformant to HTTP specs and cURL usage. if you try to launch `autocannon` with the standard syntax `-H "Content-Type: application/json" it will produce malformed HTTP headers, as in the following screenshot:

image

i will send a PR extending the existing logic to manage the standard syntax and avoid a breaking change in the usage.

phra:
see #133

### DIFF
--- a/autocannon.js
+++ b/autocannon.js
@@ -91,9 +91,14 @@ function parseArguments (argvs) {
     }
 
     argv.headers = argv.headers.reduce((obj, header) => {
-      const index = header.indexOf('=')
-      obj[header.slice(0, index)] = header.slice(index + 1)
-      return obj
+      let index
+      if (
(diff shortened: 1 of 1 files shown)


mcollina:
Good catch! Can you add a test?

phra:
@mcollina i was looking for them but i could not find any for the `autocannon.js` main file.. should i write them from scratch or can i update some existing tests?

mcollina:
Oh, don't worry about it. Let's ship this.

PR #135, opened 2018-05-22
Add badge to display install size

styfle:
This adds a badge to the `README.md` file to display the npm install size


mcollina:
Thanks for the contribution, but I don't think this is a valuable metric to evaluate a command line utility.

PR #139, opened 2018-06-20
Duration in milliseconds

baudev:
It could be interesting having every duration in milliseconds. 
I see any advantage having duration in seconds except losing precision.

There are some more steps before merging this PR (if only you would):

- [ ] Edit tests
- [ ] Edit documentation

### DIFF
--- a/lib/run.js
+++ b/lib/run.js
@@ -245,7 +245,7 @@ function run (opts, cb) {
     }
 
     if (typeof opts.duration === 'string') {
-      if (/[a-zA-Z]/.exec(opts.duration)) opts.duration = timestring(opts.duration) * 1000
+      if (/[a-zA-Z]/.exec(opts.duration)) opts.duration = timestring(opts.duration, 'ms')
       else opts.duration = Number(opts.duration.trim()) * 1000
     }
 
--- a/lib/run.js
+++ b/lib/run.js
@@ -100,7 +100,7 @@ function run (opts, cb) {
   if (!amount) {
     var stopTimer = setTimeout(() => {
       stop = true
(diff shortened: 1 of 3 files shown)


mcollina:
unit tests are failing, would you mine updating them ad well?

Il giorno mer 20 giu 2018 alle 08:26 baudev <notifications@github.com> ha
scritto:

baudev:
It's what I've planned! But when I'll have time.

lib/run.js
```diff
@@ -245,8 +245,8 @@ function run (opts, cb) {
     }
 
     if (typeof opts.duration === 'string') {
-      if (/[a-zA-Z]/.exec(opts.duration)) opts.duration = timestring(opts.duration)
-      else opts.duration = Number(opts.duration.trim())
+      if (/[a-zA-Z]/.exec(opts.duration)) opts.duration = timestring(opts.duration) * 1000
```
thekemkid: Would you please change this `timestring(opts.duration) * 1000` to `timestring(opts.duration, 'ms')` if you can, see: https://www.npmjs.com/package/timestring#return-time-value

lib/run.js
```diff
@@ -245,8 +245,8 @@ function run (opts, cb) {
...
+      else opts.duration = Number(opts.duration.trim()) * 1000
```
thekemkid: should this be `* 1000`? surely if duration was a string it should be in ms, too?

baudev:
@thekemkid thanks for this more beautiful way of doing it.

mcollina:
Tests still need updating!

mcollina:
Closing for no activity, feel free to reopen!

PR #144, opened 2018-07-18
List percentiles in output data, fixes #138

### LINKED ISSUE #138: Misleading latency output
This tool outputs latency averages and stddev which makes a lot of people assume Normal distribution that latencies usually don't have.

```
$ autocannon -c 1 -r 1 http://example.com
Running 10s test @ http://example.com
1 connections

Stat         Avg     Stdev  Max     
Latency (ms) 242.4   351.36 1296.58 
Req/Sec      1       0.45   2       
Bytes/Sec    1.57 kB 701 B  3.19 kB 

10 requests in 10s, 15.8 kB read
```

https://bravenewgeek.com/everything-you-know-about-latency-is-wrong
https://www.elastic.co/blog/averages-can-dangerous-use-percentile

### LINKED ISSUE #143: Inform about the number of observations
To compute a confidence interval or determine if there is a statistical difference between two versions, at least three pieces of information is required:

* the mean
* the standard deviation (or variance)
* the number of observations, that the mean and standard deviation were sampled from.

The number of observations is not really clear from the output.

In a previous conversation @mcollina wrote:

> The approach that autocannon takes is that it samples once per second the number of requests that happened within that second (https://github.com/mcollina/autocannon/blob/master/lib/run.js#L119-L122). So, it's 30 samples.

It is not really clear to me how that data is aggregated. Or how it relates to 30 samples.

goto-bus-stop:
This changes the Latency output to Avg / 90% / 99% / Max. The Req/Sec
and Bytes/Sec is unchanged.

```
Running 10s test @ https://wlk.yt
10 connections

Stat         Avg    90% 99% Max
Latency (ms) 255.41 337 715 987.17

Stat      Avg     Stdev  Max
Req/Sec   38.4    5.97   46
Bytes/Sec 6.14 MB 954 kB 7.37 MB

384 requests in 10s, 61.5 MB read
```

I'm no stats expert but from #138 my understanding is that stdev is fine
for the other two metrics so I just split the tables. (Maybe latency
could use a little cell margin to make it easier to read the percentile
figures.)

Lmk if the percentiles should also be used for the req and bytes/sec

Fixes #138 
Fixes #143


mcollina:
> Lmk if the percentiles should also be used for the req and bytes/sec

I think so, yes.

This would require an update also on the format for the programmatic API, as the percentiles are not included in result.requests and the others. These are important data.

For all, I think we should add: 2.75% 50% 97.5% percentiles. See https://github.com/mcollina/autocannon/issues/143#issuecomment-400651631.

Something like:

Stat  | P2.75 | P50 | P97.5 | P99 | Avg  | Stdev | Max

Alternatively:

Stat  | P2.75 | P50 | P97.5 | P99

@AndreasMadsen what do you think?

AndreasMadsen:
For sure the standard deviation is pretty useless for humans. You properly want the std error. The P2.75 and P97.5 are fine too, and properly prefered because the distributed is likely to be screwed (details below). For the same reason it is a good idea to include P50 (median). – However, the standard deviation/variance and number of observations is the only reasonable summaries if you want to calculate a statistical signifiance later.

Math details:

Time (latency) is very often [gamma distributed](https://en.wikipedia.org/wiki/Gamma_distribution), this is a screwed distrubution which means that the "mean" is unlikely to be close to the "median". One is **not** more correct than the other (I don't know why so many belive this), they are just different ways of summarizing something. As with all summaries, something will be lost. The different is just in what will be lost. The book "How to lie with statistics" is pretty good at explaining this is layman's terms.

Regarding the difference between latency, and Req/Sec or Bytes/Sec. The only diffrence is that those should have a reciprocal gamma distribution. This turns out to be the same as the [inverse-gamma distribution](https://en.wikipedia.org/wiki/Inverse-gamma_distribution) ([proof](http://www.math.wm.edu/~leemis/chart/UDR/PDFs/GammaInvertedgamma.pdf)). In all honesty, there is much difference in terms of how well they summarize using typical statistics, so whatever you argue for latency should apply for Req/Sec and Bytes/Sec (this is where you can prove me wrong).

lib/run.js
```diff
@@ -13,6 +13,10 @@ const reInterval = require('reinterval')
 const histAsObj = histUtil.histAsObj
 const addPercentiles = histUtil.addPercentiles
 
+// TODO upstream to hdr-histogram-percentiles-obj?
+histUtil.percentiles.push(2.5, 97.5)
+histUtil.percentiles.sort((a, b) => a - b)
```
mcollina: I think those needs to be upstreamed.

lib/run.js
```diff
@@ -130,9 +134,9 @@ function run (opts, cb) {
         title: opts.title,
         url: opts.url,
         socketPath: opts.socketPath,
-        requests: histAsObj(requests, totalCompletedRequests),
+        requests: addPercentiles(requests, histAsObj(requests, totalCompletedRequests)),
         latency: addPercentiles(latencies, histAsObj(latencies)),
-        throughput: histAsObj(throughput, totalBytes),
+        throughput: addPercentiles(throughput, histAsObj(throughput, totalBytes)),
```
mcollina: Can you please add some tests that verify that those are there?

lib/run.js
goto-bus-stop: Aye: https://github.com/thekemkid/hdr-histogram-percentiles-obj/pull/2 May have to change the 2.5% name (as mentioned in the PR)

lib/progressTracker.js
```diff
@@ -175,6 +177,12 @@ function asColor (colorise, row) {
 
 function asBytes (stat) {
   const result = Object.create(stat)
+
+  // Percentiles
+  Object.keys(stat).filter(k => /^p\d+$/.test(k)).forEach((k) => {
+    result[k] = prettyBytes(stat[k])
+  })
```
mcollina: This is not matching 97_5 and 2_5 percentiles.

lib/progressTracker.js
```diff
@@ -87,9 +87,7 @@ function track (instance, opts) {
         paddingRight: 1
       },
       drawHorizontalLine: () => false
-    })
-
-    logToStream(out)
+    }))
 
     if (opts.renderLatencyTable) {
```
mcollina: This will need an update as well, in conjunction with https://github.com/thekemkid/hdr-histogram-percentiles-obj/pull/2 I get:

```
$ node autocannon.js -c 100 -d 5 -l localhost:3000
Running 5s test @ http://localhost:3000
100 connections

Stat         2.5%    50%     97.5%   99%     Avg     Stdev   Max
Latency (ms) 3       4       8       11      4.66    2.19    93.68
Req/Sec      17503   19823   20623   20623   19384   1105.74 20619
Bytes/Sec    2883583 3.28 MB 3407871 3.41 MB 3.16 MB 196 kB  3.38 MB

Percentile      Latency (ms)
2.5
50              4
75              5
90              6
97.5
99              11
99.9
99.99
99.999

97k requests in 5s, 15.9 MB read
```

goto-bus-stop:
(CI failures are because the old hdr-histogram-percentiles-obj is still used)

mcollina:
Can you please update the docs as well?

AndreasMadsen:
I would like to point out that we want Latency to be low but  Req/Sec and Bytes/Sec to be high. An asymmetric summary thus doesn't make sense.

Right now, you are including max which makes sense for latency. But properly we are mostly interested in the min for Req/sec and Bytes/sec. Similarly, you are including the 99% quantile. However, the 1% quantile is properly more interesting for Req/sec and Bytes/sec.

mcollina:
What’s your recommendation @AndreasMadsen?

goto-bus-stop:
Does something along these lines make more sense?

```
Stat         2.5%    50%     97.5% 99%    Avg     Stdev   Max
Latency (ms) 20      46      131   171    51.14   26.72   215.32
             1%      2.5%    50%   97.5%  Avg     Stdev   Min
Req/Sec      158     158     172   314    193.2   44.37   158
Bytes/Sec    81.9 kB 81.9 kB 86 kB 164 kB 97.3 kB 23.6 kB 78.5 kB
```

AndreasMadsen:
@goto-bus-stop Statistically it makes sense. Although, I'm not sure how easy it is to read like that.

goto-bus-stop:
yeah. could also just add all of them but then you have to filter them when reading. in a terminal the headers have colours which helps a little. i'll play with it to get at least that info on the screen then.

goto-bus-stop:
It's not terrible I think as two white space separated tables:

image

What do y'all think?

It's a bit weird to have the Min column at the end, but I'm not sure flipping it would be better.

mafintosh:
@goto-bus-stop I think the last UI snippet is really good!

goto-bus-stop:
looks like the appveyor failures were fixed by increasing the timeout!

lib/progressTracker.js
```diff
@@ -75,27 +75,31 @@ function track (instance, opts) {
...
         paddingLeft: 0,
         paddingRight: 1
       },
       drawHorizontalLine: () => false
-    })
+    }
 
-    logToStream(out)
+    logToStream(table([
+      asColor(chalk.cyan, ['Stat', '2.5%', '50%', '97.5%', '99%', 'Avg', 'Stdev', 'Max']),
```
BridgeAR: As a suggestion: rename `50%` to `Median` in both cases.

BridgeAR:
There is just one thing I miss: a description how to read this properly. That is likely out of scope for this PR but I suggest to create a small static webpage that describes everything in more detail - also how to read the statistical data and to add a link to that webpage next to the output (e.g., `Detailed explanation: https://nearform.com/autocannon`).
It would also be nice if that link would pass through the gathered stats so the page would be able to explain everything with that data. That could be done by asking the user if he wants to have a detailed explanation and to upload the data to nearform in that case.

BridgeAR:
I just fear most people do not read the README ;-)

But it's still a good point to add a description to it!

mcollina:
@goto-bus-stop could you add that to the README?

goto-bus-stop:
This was merged as part of #145 (I merged this into that branch to make tests pass reliably). I'll try write something about how to read it this week :)

PR #145, opened 2018-07-18
--on-port

goto-bus-stop:
Do `autocannon --on-port /api/expensive-route -- node server.js` to cannonise a script once it's booted up.

This works by delaying the `run()` call until the script has started listening on a port, and then starting autocannon targeting `http://localhost:${THAT_PORT}/api/expensive-route`. Once over the script gets a SIGINT like in 0x.

There are a few things I'm not certain about:

 - ~~0x implements the child process spawning inside its Node API, this patch only implements it in the CLI. This is quite a bit easier to do because all of the `run()` setup doesn't have to be async this way, but maybe there is a use case for this functionality in the Node API too?~~
 - ~~I'm just assuming targeting 'localhost:PORT' is the right thing to do, but maybe it's not? I'm not sure what else it could be, and users can specify a full URL if they need rather than just a path, so it shouldn't be a problem in most cases.~~
 - ~~Should probably actually use [nitm](https://www.npmjs.com/package/nitm) instead~~

Same URL disclaimer as #147:

> It uses the URL API which was backported to Node 6.13.0, if lower
versions of Node 6 should be supported I can change it to use
`url.parse` and `url.format` instead.


mcollina:
> 0x implements the child process spawning inside its Node API, this patch only implements it in the CLI. This is quite a bit easier to do because all of the run() setup doesn't have to be async this way, but maybe there is a use case for this functionality in the Node API too?

I think this is only a CLI feature. However, I'm good with placing it in exposing it as a reusable function, similar to what we do for `track`: https://github.com/mcollina/autocannon/blob/master/autocannon.js#L13. I don't think this should go into `run`.

> I'm just assuming targeting 'localhost:PORT' is the right thing to do, but maybe it's not? I'm not sure what else it could be, and users can specify a full URL if they need rather than just a path, so it shouldn't be a problem in most cases.

That's a safe assumption, yes.

> Should probably actually use nitm instead

Go for it!

mcollina:
(CI is failing)

help.txt
```diff
@@ -14,6 +14,8 @@ Available options:
         The amount of requests to make before exiting the benchmark. If set, duration is ignored.
   -S/--socketPath
         A path to a Unix Domain Socket or a Windows Named Pipe. A URL is still required in order to send the correct Host header and path.
+  --on-port PATH
+        Start the command listed after -- on the command line. When it starts listening on a port, start sending requests to that port with the given URL path.
```
mcollina: We might want to move `PATH` out from this option. See #146.

goto-bus-stop: yeah when i wrote that line my understanding of how it would work was wrong. it's a boolean flag in the code and PATH is just the URL. will update :+1:

goto-bus-stop:
Hmm looking at the CI failures I would guess that `latency.average` ends up at 0 so the `t.ok()` check fails. I'll change it to `t.type`

mcollina:
Can you please update the docs as well?

autocannon.js
```diff
@@ -139,9 +151,37 @@ function start (argv) {
     return
   }
 
+  if (argv.onPort) {
+    const proc = nitm(
+      ['-r', require.resolve('./lib/detectPort')],
```
mafintosh: You should use NODE_PATH instead, as this will fail on windows with usernames with spaces in them (which lots of people have there).

Move detectPort.js to  another folder, fx `./injects` and set `process.env.NODE_PATH = path.join(__dirname, '/injects')` and use `['-r', 'detectPort']` to work around that.

lib/detectPort.js
```diff
@@ -0,0 +1,17 @@
...
+  const port = Buffer.from(addr.port + '')
+  fs.writeSync(3, port, 0, port.length)
+})
+
+// `nitm` catches the SIGINT so we write it to a file descriptor instead
+const stream = fs.createReadStream('/', { fd: 3 })
+process.on('exit', () => stream.close())
+stream.once('data', (chunk) => {
+  if (chunk.toString() === 'SIGINT') process.exit(0)
+})
```
mafintosh: Maybe we should unref the stream here? Just incase the server closes gracefully and we don't wanna keep it open.

mafintosh:
Looks good other than my comments :+1:

autocannon.js
goto-bus-stop: I took it from https://github.com/davidmarkclements/0x/blob/3611fc7f7c04ee741daeda33c972bb26cd7bdd0b/platform/v8.js#L34, so we should also change it there then. Does windows not support quoted arguments? :scream:

lib/detectPort.js
goto-bus-stop: Good call!

goto-bus-stop: Hmm actually not sure how to do this, fs.ReadStream doesn't have an unref(). autocannon will send the SIGINT when it's done anyway so currently the process just keeps running for those additional seconds. Ideally the child process would exit when the server closes and autocannon would also stop, I guess?

goto-bus-stop:
Oh, on-net-listen needs async_hooks, so this actually can't support Node 6 at all. Added a check for that too. The other CI failures are fixed by #144

mafintosh:
@goto-bus-stop oops, sorry about that. just published.

goto-bus-stop:
merci merci!

it seems like on windows there is no guarantee that the additional pipe is opened at fd=3: https://github.com/nodejs/node/issues/14046#issuecomment-385333737

prolly need to use a temp socket instead

mafintosh:
@goto-bus-stop interesting about the fd stuff. we acutally just use fd 3 in the other modules and it seems to work quite well on windows (maybe just by chance!). Would be good to put these pseudo signals in a module also so we can test it independently etc

goto-bus-stop:
hmm I tried 0x's --on-port and it didn't work either in a Win7 or Win10 
VM. it seems like the stdout redir did work there tho, actually, which 
also relies on fd 3. 0x doesn't use nitm yet, could it be that the extra 
pipe isn't passed through the proxy cmds that nitm uses on windows?

mafintosh:
@goto-bus-stop really? it's works for me consistently on windows. are you sure you are using the latest version of 0x?

goto-bus-stop:
I'm using git master :/ I'll take a closer look next week at what 
exactly is going wrong … it sounds like it might be on my end then? tho 
this is a clean install with just Win10 + Git + Node and using the 0x 
rest-api example from PowerShell.

goto-bus-stop:
omg, the onPort test works when run as `.\node_modules\.bin\tap test\onPort.test.js`, but not when run through `npm test`… :face_with_head_bandage:

goto-bus-stop:
After discussing with @mafintosh i removed nitm in favour of using NODE_OPTIONS, which is supported in all versions with async_hooks. on-net-listen requires async_hooks, so it was already required for --on-port.

Technically it's possible now to drop the socket stuff and just use fd 3 like 0x does, but I'm not entirely sure if fd 3 is really guaranteed or if it just usually ends up working, so the socket way may still be more reliable...

I think this is ready, but I merged #144 into this to fix flaky tests on this branch which ruins the diff view, so it would be easiest if #144 was merged to master first and this was PR rebased.

mafintosh:
LGTM, great work @goto-bus-stop

mcollina:
Great work!

PR #147, opened 2018-07-20
Target localhost by default if $PORT is set, closes #146.

### LINKED ISSUE #146: Better support clinic and 0x integration
We should be able to run:

```
0x --on-port 'autocannon /hello' -- node index.js
```

The `--on-port` on 0x adds a `PORT` env variable, and we can assume `localhost` as the address.
It should error if the `PORT` env variable is not available.

goto-bus-stop:
If the `$PORT` env variable is set, the URL defaults to
`localhost:PORT/path`. You can pass a bare path for the URL.

This allows doing:

    0x --on-port 'autocannon /path' -- node server.js

When an invalid URL is given, autocannon now logs an error and exits.
(`/path` without a `$PORT` is an invalid URL.)

It uses the URL API which was backported to Node 6.13.0, if lower
versions of Node 6 should be supported I can change it to use
`url.parse` and `url.format` instead.

### DIFF
--- a/autocannon.js
+++ b/autocannon.js
@@ -5,6 +5,7 @@
 const minimist = require('minimist')
 const fs = require('fs')
 const path = require('path')
+const URL = require('url').URL
 const help = fs.readFileSync(path.join(__dirname, 'help.txt'), 'utf8')
 const run = require('./lib/run')
 const track = require('./lib/progressTracker')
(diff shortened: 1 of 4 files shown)


autocannon.js
```diff
@@ -5,6 +5,7 @@
 const minimist = require('minimist')
 const fs = require('fs')
 const path = require('path')
+const URL = require('url').URL
```
mcollina: Can you throw an error if `URL` is not available?

mcollina:
🎉 !

PR #149, opened 2018-07-21
Updated docs for input and body options

zerodom30:


zerodom30:
Then this detail would make more sense as a side `Note`. Going to address the build issues as well.

mcollina:
Thanks!

PR #155, opened 2018-08-22
Explain the CLI table output in the readme

goto-bus-stop:
As requested in #144

I just mentioned what the different fields mean in a pretty basic sense, and what values users should be hoping for (high/low). I don't know much of anything about the statistical significance so I didn't add anything about that.

I included the -l latency table for completeness but it seems less useful now that the small default latency table already lists key percentiles...

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,53 @@ Available options:
         Print this menu.
 ```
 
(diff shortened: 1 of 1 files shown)


AndreasMadsen:
LGTM

PR #157, opened 2018-09-03
fix argv support in windows cmd.exe

### LINKED ISSUE #153: All responses are non 2xx with method POST in Hapi.js
I made same post in hapi/disscuss (https://github.com/hapijs/discuss/issues/705) repository, since this only apeared with Hapi framework. But then i realized that with GET method it works fine, but POST fails.
I added header property in the call and the problem still persists.

hapierr2

mcollina:
Fixes #153.


mcollina:
cc @mafintosh

autocannon.js
```diff
@@ -238,5 +239,7 @@ function runTracker (argv, ondone) {
 }
 
 if (require.main === module) {
-  start(parseArguments(process.argv.slice(2)))
+  const argv = crossArgv(process.argv.slice(2))
+  console.log(argv)
```
mafintosh: Did you mean to log this?

mcollina: actually no.

ccalvarez:
Hi.  I'm using `autocannon@3.2.0` today, and I'm still needing the \ escape character.  

Operating system:  Windows 10 Pro Version 1709, Build 16299.847


image

Thank you.

PR #164, opened 2018-09-30
Fix issue with host header

Tuizi:
Add the possibility to set the `host` header.

### DIFF
--- a/lib/httpRequestBuilder.js
+++ b/lib/httpRequestBuilder.js
@@ -37,7 +37,7 @@ function requestBuilder (defaults) {
     const headers = reqData.headers
     const body = reqData.body
 
-    let host = reqData.host
+    let host = reqData.headers.host || reqData.host
     if (!host) {
       const hostname = reqData.hostname
       const port = reqData.port
--- a/package.json
+++ b/package.json
@@ -7,7 +7,8 @@
     "autocannon": "autocannon.js"
   },
   "scripts": {
-    "test": "standard && tap --timeout 45 test/*.test.js"
+    "test": "standard && tap --timeout 45 test/*.test.js",
+    "standard:fix": "standard --fix"
   },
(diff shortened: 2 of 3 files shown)


mcollina:
Good work! Can you update the docs and add a cli argument to set this as well?

Tuizi:
There is no specific argument or doc needed I guess. It is just a "bug fix".
Before the `host` header was ignored no it is taken in consideration.

What do you think?

lib/httpRequestBuilder.js
```diff
@@ -37,7 +37,7 @@ function requestBuilder (defaults) {
     const headers = reqData.headers
     const body = reqData.body
 
-    let host = reqData.host
+    let host = reqData.headers.host || reqData.host
```
silverwind, line 40: This does seem to fail when the header is given as `-H Host=value` on the CLI, e.g. it does not work case-insensitively like HTTP headers are supposed to be.

mcollina: wow, this was a long time ago, would you open a fresh issue or a PR to fix?

silverwind: https://github.com/mcollina/autocannon/pull/463

exoticknight: I use
```bash
autocannon -H "host=xxx.com" "http://127.0.0.1:8000"
```
it seems that it just replace 127.0.0.1 with xxx.com in url which is not what i intend

```
Request {
  bodyUsed: false,
  headers: Headers { connection: "keep-alive", host: "xxx.com, xxx.com" },
  method: "GET",
  redirect: undefined,
  url: "http://xxx.com/"
}
```

Deno serve
```ts
Deno.serve((req) => {
  console.log(req)
  return new Response('a')
})
```

PR #166, opened 2018-10-12
fix connectionRate typo in README

songawee:
This tool is amazing! Thanks for providing it to the community :). Noticed a small typo and thought I'd push a fix upstream. Only misspelled `connection` 1/32 times which is much better than myself 👍


mcollina:
Thanks for catching this!

Issue #169, opened 2018-10-16
Use native Object.assign instead of xtend


jimmywarting:
[xtend](https://github.com/Raynos/xtend) can easily be replaced with [Object.assign](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign)

```js
// immutable
Object.assign({}, a, b)

// mutable
Object.assign(a, b)
```

mcollina:
Can you please send a PR?

Il giorno mar 16 ott 2018 alle 16:21 Steven <notifications@github.com> ha
scritto:

thekemkid:
We should do it. When we wrote

thekemkid:
Closed via #173

PR #173, opened 2018-10-24
Removed xtend and shallow-clone dependencies

Eomm:
Hi,
I have implemented #169 and I have removed also the `shallow-clone` dependency because was used only to clone objects and also that implementation was with `Object.assign`.

Hope this help
Regards


thekemkid:
Thank you 👏

Issue #177, opened 2018-11-26
Cannot read property 'tick' of undefined


Youjingyu:
when i use `autocannon` with the following command:
```bash
autocannon -c 500 -a 5000  http://10.222.22.67:8334/getConfig\?pageID\=1002
```
throw error:
```bash
TypeError: Cannot read property 'tick' of undefined
    at EventEmitter.instance.on (/usr/local/lib/node_modules/autocannon/lib/progressTracker.js:67:57)
```
When I comment this line, it was fine:
```javascript
      instance.on('response', () => { amountProgressBar.tick() })
      // autocannon/lib/progressTracker.js:67:57
      // comment this line
      // instance.on('reqError', () => { amountProgressBar.tick() })
      instance.on('done', () => { amountProgressBar.tick(iOpts.amount - 1) })
      process.once('SIGINT', () => { amountProgressBar.tick(iOpts.amount - 1) })
```

mcollina:
which version of Node/Operating System are you using?

mcollina:
Can you upload an example server that reproduces the behavior?

Youjingyu:
@mcollina 
Node: v8.10.0
Operating System: macos high sierra 10.13.6
autocannon: v3.1.0
an example server that reproduces the behavior: https://github.com/Youjingyu/test-autocannon

Youjingyu:
thanks, it was fixed [#178](https://github.com/mcollina/autocannon/pull/178)

PR #178, opened 2018-11-27
Initialize things only after start is emitted

mcollina:
Fixes https://github.com/mcollina/autocannon/issues/177.

@Youjingyu  can you please verify?


Youjingyu:
thanks, it was fixed

PR #179, opened 2018-11-29
remove table, use cli-table3

mcollina:
Currently we are broken as table uses some ajv internals.

After digging a bit through the http//npm.im/table code, I think we should switch to http://npm.im/cli-table3, which is also used by npm.

I also added a proper table, mainly because it make things easier to read with the new system with two tables.


mcollina:
See https://github.com/epoberezkin/ajv/issues/889 for more details.

PR #193, opened 2019-04-04
Check to prevent infinite loop when amount is greater than connections

CharlieTheProgrammer:
First time submitting a PR ever. I tried adding to checkOptsForErrors() function, but that just resulted in about 70 failed tests, adding the throw was **a lot** easier.


CharlieTheProgrammer:
I looked at httpClient.test.js, but I'm not quite sure how to properly execute an instance of a client such that the test catches the throw.

PR #204, opened 2019-05-17
Drop v6 add v12

thekemkid:


mcollina:
Would you mind bumping the deps as well? I think tap can be updated to v13.

thekemkid:
The build was broken when I bumped tap with greenkeeper. My plan was to
merge this then rebase that pr to get it green and merged 🙂

Issue #186, opened 2019-01-27
Add promise api


vjpr:

mcollina:
Would you like to help and send a PR?

banyudu:
Should the promise API introduce breaking changes or just add new export member?

Is this ok? 

```javascript
module.exports.runAsync = require('util').promisify(run)
```

mcollina:
I would prefer a case where `run` returns a promise if a callback is not specified.

banyudu:
> I would prefer a case where `run` returns a promise if a callback is not specified.

Did you mean 

```javascript
function run (opts, cb) {
  const cbPassedIn = (typeof cb === 'function')

  cb = cb || noop

  // add 3 lines
  if (!cbPassedIn) {
    return require('util').promisify(run)(opts, noop)
  }
```
?

It's a break change, though

mcollina:
No.

```js
function run (opts, cb) {
  const cbPassedIn = (typeof cb === 'function')

  cb = cb || noop

  // add 3 lines
  if (!cbPassedIn) {
    return new Promise((resolve, reject) => {
      run(opts, function (err, results) {
        if (err) return reject()
        resolve(results)
      })
    })
  }
```

banyudu:
Thanks for your reply.

Besides `require('util').promisify` which requires node v8, there's another breaking change:
`run` returns an EventEmitter named `tracker`, so it's a breaking change if just return a promise instead.

Test cases like
```javascript
  tracker.on('done', (result) => {
    if (ondone) ondone()
    if (argv.json) {
      console.log(JSON.stringify(result))
    }
  })
```
will fail.

mcollina:
I think the best way to solve this then is to make tracker a thenable. Essentially you can add the following method https://developer.mozilla.org/it/docs/Web/JavaScript/Reference/Global_Objects/Promise/then#Syntax to Tracker. Then you'll be able to `await` it.

banyudu:
Good idea, impressive!

ryhinchey:
just reading through this issue and https://github.com/mcollina/autocannon/pull/205.  Can this issue be closed?

PR #205, opened 2019-05-24
Feature/promise api refs #186

banyudu:
Feature: support promise API.

Modify test cases `throw if connections is greater than amount`: add a noop callback in this test case.

New test case `run promise`: copied from case `run`, use then instead of callback

### DIFF
--- a/lib/run.js
+++ b/lib/run.js
@@ -11,7 +11,11 @@ const reInterval = require('reinterval')
 const histAsObj = histUtil.histAsObj
 const addPercentiles = histUtil.addPercentiles
 
-function run (opts, cb, tracker) {
+function run (opts, cb) {
+  return _run(opts, cb)
+}
+
+function _run (opts, cb, tracker) {
   const cbPassedIn = (typeof cb === 'function')
 
   cb = cb || noop
@@ -19,13 +23,14 @@ function run (opts, cb, tracker) {
   tracker = tracker || new EE()
 
   if (!cbPassedIn && !opts.forever) {
-    tracker.promise = new Promise((resolve, reject) => {
-      run(opts, function (err, results) {
+    const promise = new Promise((resolve, reject) => {
+      _run(opts, function (err, results) {
         if (err) return reject(err)
         resolve(results)
       }, tracker)
     })
-    tracker.then = tracker.promise.then.bind(tracker.promise)
+    tracker.then = promise.then.bind(promise)
+    tracker.catch = promise.catch.bind(promise)
     return tracker
   }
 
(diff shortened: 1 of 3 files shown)


banyudu:
refs https://github.com/mcollina/autocannon/issues/186

thekemkid:
That lgtm, thank you! Minor request: Would you mind adding a small section or note into the readme too, so users know this is possible?

banyudu:
> That lgtm, thank you! Minor request: Would you mind adding a small section or note into the readme too, so users know this is possible?

Of course. Already submitted.

thekemkid:
LGTM, just gonna await @mcollina's approval before merging and releasing :)

lib/run.js
```diff
@@ -11,12 +11,23 @@ const reInterval = require('reinterval')
 const histAsObj = histUtil.histAsObj
 const addPercentiles = histUtil.addPercentiles
 
-function run (opts, cb) {
+function run (opts, cb, tracker) {
```
mcollina: Can you rename this `_run` and have the wrapper be `run` instead? I would like to avoid passing in the extra `tracker` argument, it could lead to side effects.

lib/run.js
```diff
@@ -11,12 +11,23 @@ const reInterval = require('reinterval')
...
+  tracker = tracker || new EE()
+
+  if (!cbPassedIn && !opts.forever) {
+    tracker.promise = new Promise((resolve, reject) => {
+      run(opts, function (err, results) {
+        if (err) return reject(err)
+        resolve(results)
+      }, tracker)
+    })
+    tracker.then = tracker.promise.then.bind(tracker.promise)
```
mcollina: I would also add `tracker.catch = tracker.promise.catch.bind(tracker.promise)

lib/run.js
```diff
@@ -11,12 +11,23 @@ const reInterval = require('reinterval')
...
+function run (opts, cb, tracker) {
   const cbPassedIn = (typeof cb === 'function')
 
   cb = cb || noop
 
-  const tracker = new EE()
+  tracker = tracker || new EE()
+
+  if (!cbPassedIn && !opts.forever) {
+    tracker.promise = new Promise((resolve, reject) => {
```
mcollina: do we need to keep a reference to the promise on the tracker itself? I would prefer not to.

lib/run.js
banyudu: fixed

lib/run.js
banyudu: fixed

lib/run.js
banyudu: fixed

Issue #194, opened 2019-04-18
Problem running 2 autocannons in parallel when using setupClient()


frankorama:
I have 2 files each running autocannon inside a promise.  Each file calls a different api end-point.

When I call the autocannon() method to create an instance I set the initial values of the header and body, and then change the values in an  `instance.on('response', handleResponse) method`.

Both files can run in parallel and everything works fine.

However, when I try to set the initial client header and body in the `setClient` method in file1, the body of file1 is suddenly being sent to the end-point of file2.

Can autocannon be run in parallel?  Are the instances distinct or is there shared code between the instances?

mcollina:
This sounds like a bug. Would you like to send a PR?

karpotkin:
Same issue. The more connections, the more net the module creates simultaneous requests at one time. For example, I want to create 4 connections. The first 3 connections will be sent by 1 request, and 4 connections will send 4 requests at once.

mcollina:
@karpotkin would you like to send a PR?

karpotkin:
@mcollina, ok, I will try

mcollina:
Fixed in #206.

PR #206, opened 2019-06-17
Fix autocannon running in parallel when using setupClient issue #194

karpotkin:


lib/httpClient.js
```diff
@@ -7,37 +7,42 @@ const tls = require('tls')
 const retimer = require('retimer')
 const HTTPParser = require('http-parser-js').HTTPParser
 const RequestIterator = require('./requestIterator')
+const deepCopy = require('./deepCopy')
```
mcollina: can you use http://npm.im/rfdc?

temtum: it is better not to use

karpotkin: installed https://npm.im/clone

Issue #151, opened 2018-08-08
Question: result.requests.total versus result.requests.sent


alienspaces:
Hi there,

When looking at the result object the requests object has a total and a sent property.

What is the difference?

If I set maxOverallRequests the total property will end up equalling that however the sent is always more.

Regards,

Ben

mcollina:
We stop _after_ we receive X requests while it would have sent more. Having them being equal would be a good refactoring, but if you like a challenge feel free to send a PR.

PR #207, opened 2019-06-18
Wait until all requests are executed #151

karpotkin:


lib/httpClient.js
```diff
@@ -173,10 +179,32 @@ Client.prototype.destroy = function () {
...
+        this.removeAllListeners('response')
+
+        resolve()
+      })
+    })
   }
+
+  return new Promise((resolve) => {
+    resolve()
+  })
```
mcollina: This whole module is based on the concept of callbacks, not promises. Would you mind using callbacks instead?
Also the above code _might_ leak memory in case of an error.

Would you mind adding a few comment/description on why we are doing all of this?

karpotkin: I want to know the exact number of requests sent

lib/run.js
```diff
@@ -137,80 +139,83 @@ function _run (opts, cb, tracker) {
...
+    result.requests.sent = totalRequests
+    statusCodes.forEach((code, index) => { result[(index + 1) + 'xx'] = code })
+    if (result.throughput.min === Number.MAX_SAFE_INTEGER) result.throughput.min = 0
+    if (result.latency.min === Number.MAX_SAFE_INTEGER) result.latency.min = 0
+
+    tracker.emit('done', result)
+    if (!opts.forever) cb(null, result)
+
+    // the restart function
+    setImmediate(() => {
```
mcollina: Can you move this closure to a top level function?
Why are you calling it with `setImmediate`?

lib/httpClient.js
```diff
@@ -169,13 +175,40 @@ Client.prototype._destroyConnection = function () {
...
+Client.prototype._waitAndDestroy = function () {
+  const waitTimer = setTimeout(() => {
+    this._requestCompleted()
+  }, this.timeout)
+
+  this.on('response', () => {
+    clearTimeout(waitTimer)
+    this._requestCompleted()
+  })
+}
```
mcollina: Can we do this in `handleTimeout` instead? We could just avoid calling `this.timeoutTicker.clear()` during `destroy()` if there are ongoing requests.

temtum:
I forgot to say that this need arises only in case of limited requests.

test/httpClient.test.js
```diff
@@ -400,6 +400,56 @@ test('client customiseRequest function overwrites the headers and body', (t) =>
...
+  })
+
+  client.on('done', () => {
+    t.equal(cnt, opts.rate + 1, 'response count match')
+    t.end()
+  })
+
+  setTimeout(() => {
+    client.destroy()
+  }, 1000)
```
mcollina: Maybe we can control this from the server instead? `setTimeout` is very brittle in tests.

karpotkin: How do you propose to do this? We need to test waiting for a response from the server when there are input parameters: `duration` and `rate`. In this case, `setTimeout` acts as a `duration`. Or it is necessary to describe tests in `run.test.js`

karpotkin:
The first time the tests did not pass on node 12. I suggest to remove the test "http client should not wait responses" because from time to time the result can be 2 instead of 1. What do you think?

mcollina:
> The first time the tests did not pass on node 12. I suggest to remove the test "http client should not wait responses" because from time to time the result can be 2 instead of 1 (I think you understand why this can happen). As for the second test, the result should always be 2. What do you think?

Go for it.

mcollina:
Would you mind adding this new option to the README as well?

karpotkin:
done

thekemkid:
This is really great work! Thank you!

I do have two pieces of feedback I'd like to see addressed - I would worry about users thinking the benchmark cli has frozen when awaiting long running responses, and I think this is complicated enough to require another test cases.

In the case `waitAllResponses` is set I would like to see a message printed to the console to state we're waiting for responses to finish before reporting stats. This will probably require emitting another event `waiting` for when the duration is reached and `waitAllResponses` is set. It might be good to report how many is being waited on as an arg of this event.

As for the test cases validation is needed that when `waitAllResponses` is set for a server not sending any responses it will still timeout and provide stat data

Thank you again!

karpotkin:
@thekemkid, good issues. I agree, I will add. @mcollina, what do think? Should I add this event to current PR?

mcollina:
> In the case waitAllResponses is set I would like to see a message printed to the console to state we're waiting for responses to finish before reporting stats. This will probably require emitting another event waiting for when the duration is reached and waitAllResponses is set. It might be good to report how many is being waited on as an arg of this event.

Currently we are not exposing the `waitAllResponses` flag to the CLI tool. Maybe we should add that and add that event as well.

> Should I add this event to current PR?

Go ahead, yes!

mcollina:
@karpotkin any updates on this?

mcollina:
Closing for no activity.

PR #211, opened 2019-07-21
Avoid maxListenersExceededWarning on process when using the tracker API

lmammino:
This PR introduces a small change on the Tracker internals to make sure that the process on SIGINT event listener is cleaned up when the tracker is finished.

This can happen, for instance, when running multiple instances of autocannon in sequence to benchmark different variations of a given webserver.

The actual warning I got looks like this:

```
(node:39800) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 SIGINT listeners added to [process]. Use emitter.setMaxListeners() to increase limit
```

## TODO

 - [ ] Add more coverage to new changes
 - [x] Make tests windows compatible (currently failing with `C:\dev\null does not exist`)
 - [ ] single listener to SIGINT in the runTracker
 - [ ] add a method in the progress tracker that we’d call to during SIGINT

### DIFF
--- a/lib/progressTracker.js
+++ b/lib/progressTracker.js
@@ -1,5 +1,4 @@
 'use strict'
-
 const ProgressBar = require('progress')
 const Table = require('cli-table3')
 const Chalk = require('chalk')
@@ -81,7 +80,6 @@ function track (instance, opts) {
           process.removeListener('SIGINT', completeDurationProgress)
         })
         process.once('SIGINT', completeDurationProgress)
-        console.log(process._events['SIGINT'])
       } else { // amount progress bar
         instance.on('response', () => { amountProgressBar.tick() })
         instance.on('reqError', () => { amountProgressBar.tick() })
@@ -90,7 +88,6 @@ function track (instance, opts) {
           process.removeListener('SIGINT', completeAmountProgress)
         })
         process.once('SIGINT', completeAmountProgress)
-        console.log(process._events['SIGINT'])
       }
     }
   }
--- a/lib/progressTracker.js
+++ b/lib/progressTracker.js
@@ -62,32 +62,32 @@ function track (instance, opts) {
 
     addedListeners = true
 
-    function completeDurationProgress () {
-      durationProgressBar.tick(iOpts.duration - 1)
-    }
-
-    function completeAmountProgress () {
-      amountProgressBar.tick(iOpts.amount - 1)
+    function cancel () {
+      if (iOpts.amount) {
+        amountProgressBar.tick(iOpts.amount - 1)
+      } else {
+        durationProgressBar.tick(iOpts.duration - 1)
+      }
+      instance.cancel()
     }
 
     // note: Attempted to curry the functions below, but that breaks the functionality
     // as they use the scope/closure of the progress bar variables to allow them to be reset
     if (opts.outputStream.isTTY) {
+      process.once('SIGINT', cancel)
+      instance.on('done', () => process.removeListener('SIGINT', cancel))
+
       if (!iOpts.amount) { // duration progress bar
         instance.on('tick', () => { durationProgressBar.tick() })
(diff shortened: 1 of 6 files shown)


mcollina:
Can you add a unit test?

lmammino:
Seems like a good excuse to try to understand how to catch node warnings :) I'll try to do it!

lmammino:
@mcollina I added a TODO list with remaining work.

Can you pleas expand a bit on the last 2 points?

For point 3: do you mean that you want to unify the following two functions into a single one?

```javascript
function completeDurationProgress () {
  durationProgressBar.tick(iOpts.duration - 1)
}

function completeAmountProgress () {
  amountProgressBar.tick(iOpts.amount - 1)
}
```

Maybe something like this?

```javascript
function completeProgress () {
  if (iOpts.duration) {
      amountProgressBar.tick(iOpts.duration - 1)
  } else {
      amountProgressBar.tick(iOpts.amount - 1)
  }
}
```

For the 4th point: does it have to be some function that the customer can provide as part of the options, like `onSigInt` and we need to make sure this function (if specified) is called, or did you mean something else?

Thanks a lot for your support on this, autocannon is a very interesting project to work on :)

mcollina:
I think we should have tracker.terminate() or something similar to tell the tracker that we are abruptly terminating things before hand. In the runner, we add a single sigint handler that call terminate. Does it make sense?

I’m ok with the mocking.

lmammino:
@mcollina, makes totally sense. I will see how to add that. Thanks for clarifying

lmammino:
I was reading the codebase a bit yesterday and trying to understand it more.

This is my current idea to address your last comment:

When you run autocannon you get back a custom event emitter instance (also *promisified* with `then` and `catch` if not passing a callback).

I am thinking to add a user-facing `.cancel()` method to the emitter. Once called it will trigger the `canceled` event.

On the tracker side, we can simply register the `SIGINT` listener and call `currentInstance.cancel()`. In case the scrollbar is enabled we also listen for `canceled` to fill the progress bar before shutting the instance down.

@mcollina, would this implementation make sense to you?

mcollina:
Yes it will!

lmammino:
Thanks for the quick feedback. I'll try to propose the necessary changes here as soon as I have some time this week :) It will require also some changes in tests and docs!

lib/progressTracker.js
```diff
@@ -63,18 +62,32 @@ function track (instance, opts) {
...
     addedListeners = true
 
+    function cancel () {
+      if (iOpts.amount) {
+        amountProgressBar.tick(iOpts.amount - 1)
+      } else {
+        durationProgressBar.tick(iOpts.duration - 1)
+      }
+      instance.cancel()
+    }
```
mcollina: Can you export it from this module and add the SIGINT handler in the runTracker function in autocannon.js instead?

lib/progressTracker.js
```diff
@@ -63,18 +62,32 @@ function track (instance, opts) {
...
+
     // note: Attempted to curry the functions below, but that breaks the functionality
     // as they use the scope/closure of the progress bar variables to allow them to be reset
     if (opts.outputStream.isTTY) {
+      process.once('SIGINT', cancel)
```
mcollina: I would prefer not having a SIGINT handler in this file at all.

lmammino:
Please be extra picky, I don't mind at all :)
I will only proceed a bit slow on this one as I am currently very short of free time and this is probably gonna last for the entire month. I hope you don't mind about that!

lib/progressTracker.js
lmammino: Will do!

lib/progressTracker.js
lmammino: Make sense, I will do the change

mcollina:
Closing for no activity.

PR #212, opened 2019-08-16
TypeError: Cannot set property 'startTime' of undefined

dplewis:
I received the following error for `opts.pipelining` value > 1.
`this.resData.length` always returned 1 for me.
[length source](http://www.ecma-international.org/ecma-262/6.0/index.html#sec-properties-of-array-instances-length)

```
TypeError: Cannot set property 'startTime' of undefined
    at Client._doRequest (/Users/testing/Documents/Github/benchmark/node_modules/autocannon/lib/httpClient.js:146:33)
    at Client._connect (/Users/testing/Documents/Github/benchmark/node_modules/autocannon/lib/httpClient.js:135:10)
    at new Client (/Users/testing/Documents/Github/benchmark/node_modules/autocannon/lib/httpClient.js:100:8)
    at initialiseClients (/Users/testing/Documents/Github/benchmark/node_modules/autocannon/lib/run.js:214:20)
    at _run (/Users/testing/Documents/Github/benchmark/node_modules/autocannon/lib/run.js:114:3)
    at run (/Users/testing/Documents/Github/benchmark/node_modules/autocannon/lib/run.js:15:10)
    at Promise (/Users/testing/Documents/Github/benchmark/lib/autocannon.js:16:20)
```

This is my first PR, let me know if more is needed.

### DIFF
--- a/lib/httpClient.js
+++ b/lib/httpClient.js
@@ -42,13 +42,13 @@ function Client (opts) {
   // used for forcing reconnects
   this.reconnectRate = this.opts.reconnectRate
 
-  this.resData = []
+  this.resData = new Array(this.opts.pipelining)
   for (let i = 0; i < this.opts.pipelining; i++) {
-    this.resData.push({
+    this.resData[i] = {
       bytes: 0,
       headers: {},
       startTime: [0, 0]
-    })
+    }
   }
 
   // cer = current expected response
--- a/lib/httpClient.js
+++ b/lib/httpClient.js
@@ -42,13 +42,13 @@ function Client (opts) {
   // used for forcing reconnects
   this.reconnectRate = this.opts.reconnectRate
 
-  this.resData = new Array(this.opts.pipelining)
-  for (let i = 0; i < this.resData.length; i++) {
-    this.resData[i] = {
+  this.resData = []
+  for (let i = 0; i < this.opts.pipelining; i++) {
+    this.resData.push({
       bytes: 0,
       headers: {},
       startTime: [0, 0]
-    }
+    })
   }
 
   // cer = current expected response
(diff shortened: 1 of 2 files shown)


mcollina:
Would you mind adding a unit test?

dplewis:
Done

lib/httpClient.js
```diff
@@ -42,13 +42,13 @@ function Client (opts) {
...
-  for (let i = 0; i < this.resData.length; i++) {
-    this.resData[i] = {
+  this.resData = []
+  for (let i = 0; i < this.opts.pipelining; i++) {
+    this.resData.push({
       bytes: 0,
       headers: {},
       startTime: [0, 0]
-    }
+    })
```
mcollina: This array is pre-allocated as a memory optimization. Maybe we should clean it up and shrinked when destroy() is called.

dplewis: That might be true for c/c++. Here is a [benchmark](http://jsperf.com/literal-vs-new-23). Literal [] is faster than new.

I'll let you take a stab at it. I got this error on node v10.14 and 10.16.

`new Array(10) => ['10'] instead of [](10)`

mcollina: That's not what I mean. The difference is in `.push()`, not on creation.

dplewis: Sorry for the confusion. I was looking at speed instead of memory.

Would this be better?
```
this.resData = new Array(this.opts.pipelining)
for (let i = 0; i < this.opts.pipelining; i += 1) {
  this.resData[i] = {
    bytes: 0,
    headers: {},
    startTime: [0, 0]
  }
}
```

mcollina:
That’d be better, yes!

Il giorno mer 21 ago 2019 alle 17:51 Diamond Lewis <notifications@github.com>
ha scritto:

dplewis:
Done

mcollina:
Thanks!

PR #217, opened 2019-09-18
add option to print errors

yoshuawuyts:
This patch adds ` --debug` flag that prints connection errors to `stderr`.

I've been messing around with HTTP parsers recently, and autocannon is a great way to test how well a server works. Being able to see errors are happening is usually a good indicator things are broken. And having a way to print out *what* those errors are is really convenient. So hence this patch.

Hope this is useful. Thanks!

### DIFF
--- a/lib/run.js
+++ b/lib/run.js
@@ -241,6 +241,7 @@ function _run (opts, cb, tracker) {
     function onError (error) {
       for (let i = 0; i < opts.pipelining; i++) tracker.emit('reqError', error)
       errors++
+      if (opts.debug) console.error(error)
       if (opts.bailout && errors >= opts.bailout) stop = true
     }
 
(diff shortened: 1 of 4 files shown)


yoshuawuyts:
@mcollina added a new file for debug tests, asserting the the option is set when passed. Thanks for the review; I hope this is good!

lib/run.js
```diff
@@ -241,6 +241,7 @@ function _run (opts, cb, tracker) {
     function onError (error) {
       for (let i = 0; i < opts.pipelining; i++) tracker.emit('reqError', error)
       errors++
+      if (opts.debug) console.error(error)
```
mcollina: I would love to have a test that requires this and mocks `console.error` to verify things gets printed out.

PR #220, opened 2019-10-06
adds support for HTTP basic authentication

### LINKED ISSUE #213: Supporting Basic HTTP auth in URL
Hi 👋

I have a staging server, which is protected with [basic HTTP authentication](https://en.wikipedia.org/wiki/Basic_access_authentication) to avoid accidental indexing by search engines. For manual testing, I simply enter login and password in the browser and for tools like [HTTPie](https://httpie.org/) or [Lighthouse](https://developers.google.com/web/tools/lighthouse/) I prepend my login and password to the URL like this:

```bash
http https://login:password@staging.example.com/
lighthouse --view https://login:password@staging.example.com/
```

All works – the response is 200.

However, `autocannon` seems to ignore my auth credentials written the same way:

```bash
autocannon https://login:password@staging.example.com/
```
```
Running 10s test @ https://login:password@staging.example.com/
10 connections

┌─────────┬───────┬───────┬────────┬────────┬─────────┬──────────┬───────────┐
│ Stat    │ 2.5%  │ 50%   │ 97.5%  │ 99%    │ Avg     │ Stdev    │ Max       │
├─────────┼───────┼───────┼────────┼────────┼─────────┼──────────┼───────────┤
│ Latency │ 77 ms │ 89 ms │ 117 ms │ 144 ms │ 93.8 ms │ 30.49 ms │ 394.45 ms │
└─────────┴───────┴───────┴────────┴────────┴─────────┴──────────┴───────────┘
┌───────────┬─────────┬─────────┬─────────┬─────────┬───────┬─────────┬─────────┐
│ Stat      │ 1%      │ 2.5%    │ 50%     │ 97.5%   │ Avg   │ Stdev   │ Min     │
├───────────┼─────────┼─────────┼─────────┼─────────┼───────┼─────────┼─────────┤
│ Req/Sec   │ 71      │ 71      │ 110     │ 117     │ 105.6 │ 12.54   │ 71      │
├───────────┼─────────┼─────────┼─────────┼─────────┼───────┼─────────┼─────────┤
│ Bytes/Sec │ 12.1 kB │ 12.1 kB │ 18.7 kB │ 19.9 kB │ 18 kB │ 2.13 kB │ 12.1 kB │
└───────────┴─────────┴─────────┴─────────┴─────────┴───────┴─────────┴─────────┘

Req/Bytes counts sampled once per second.

0 2xx responses, 1056 non 2xx responses
1k requests in 10.05s, 180 kB read
```

↑ `1056 non 2xx responses` suggests that HTTP auth was ignored and `401 Unauthorized` was returned in all cases. Could autocannon be improved so that it worked just like other tools with regards to basic auth?

gillesdemey:
fixes #213

### DIFF
--- a/lib/httpRequestBuilder.js
+++ b/lib/httpRequestBuilder.js
@@ -43,6 +43,11 @@ function requestBuilder (defaults) {
       host = hostname + ':' + port
     }
 
+    if (reqData.auth) {
+      const encodedAuth = Buffer.from(reqData.auth).toString('base64')
+      headers.Authorization = `Basic ${encodedAuth}`
+    }
+
     if (methods.indexOf(method) < 0) {
       throw new Error(`${method} HTTP method is not supported`)
     }
--- a/lib/httpClient.js
+++ b/lib/httpClient.js
@@ -22,6 +22,7 @@ function Client (opts) {
   this.timeout = (this.opts.timeout || 10) * 1000
   this.ipc = !!this.opts.socketPath
   this.secure = this.opts.protocol === 'https:'
+  this.auth = this.opts.auth || null
 
   if (this.secure && this.opts.port === 80) {
     this.opts.port = 443
--- a/test/basic-auth.test.js
+++ b/test/basic-auth.test.js
@@ -1,0 +1,54 @@
+'use strict'
+
+const t = require('tap')
+const split = require('split2')
+const path = require('path')
+const childProcess = require('child_process')
+const helper = require('./helper')
+
+const lines = [
+  /Running 1s test @ .*$/,
+  /10 connections.*$/,
+  /$/,
+  /.*/,
+  /Stat.*2\.5%.*50%.*97\.5%.*99%.*Avg.*Stdev.*Max.*$/,
+  /.*/,
+  /Latency.*$/,
+  /$/,
+  /.*/,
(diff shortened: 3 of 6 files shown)


lib/httpRequestBuilder.js
```diff
@@ -43,6 +43,11 @@ function requestBuilder (defaults) {
       host = hostname + ':' + port
     }
 
+    if (reqData.auth) {
+      const encodedAuth = Buffer.from(reqData.auth).toString('base64')
```
mcollina: I think it would be better to parse these from the URL instead, e.g. `http://matteo:collina@localhost:3000` will generate `Buffer.from(matteo:collina).toString('base64')`

gillesdemey: I'm not quite sure what you mean, afaik the URL is parsed here https://github.com/mcollina/autocannon/blob/master/lib/run.js#L79 and passed to the `Client` constructor [here](https://github.com/mcollina/autocannon/blob/master/lib/run.js#L215) so it's already parsed from the URL :)

mcollina: Oh, would you mind adding a full test that verifies it all works as expectedv

mcollina:
Any updates on this one?

gillesdemey:
I've just added an end-to-end test and also tested the negative case locally where the test fails if no credentials were passed via the URL.

PR #222, opened 2019-10-15
fix host header being sent twice

dound:
If the user specifies a custom host header, then the request will be sent with _two_ (identical) host headers. This is unnecessary, and breaks some systems (e.g., Google Kubernetes Engine fails to route inbound HTTP requests which include more than one host header).


mcollina:
Would you mind to add a unit test?

mcollina:
Close for no activity.

Issue #160, opened 2018-09-19
Request multipart/form-data


mrbatista:
Does the tool support requests with multipart/form-data?

mcollina:
You would need to encode that manually, using something like https://www.npmjs.com/package/form-data. It would be fantastic to add the capability of doing this to autocannon itself.

thekemkid:
@mrbatista Would this be something you would be interested in adding support for as a PR?

hiepph:
Hi @mcollina can you show me an example of how to encode using `form-data`. For instance, I want to parse a body `multipart/form-data` with key `image` and value is a local image file.

mcollina:
Unfortunately I don't have much time. The readme of form-data has a lot of examples: https://www.npmjs.com/package/form-data.

dnlup:
Here's an example I have used, I hope it can be useful:
```
function bench({ url, file, connections, requests }) {
    return new Promise((resolve, reject) => {
        const buffer = readFileSync(file);
        const form = new FormData();
        form.append('file', buffer, {
            filename: basename(file),
            contentType: 'application/pdf',
        });
        const instance = autocannon(
            {
                title: `Convert from file upload with concurrency ${connections} | ${requests} Requests`,
                url: `${url.origin}/convert`,
                method: 'POST',
                headers: {
                    ...form.getHeaders(),
                },
                body: form.getBuffer(),
                connections: connections,
                amount: requests,
                timeout: 30,
            },
            (error, result) => {
                process.removeAllListeners('SIGINT');
                return error ? reject(error) : resolve(result);
            }
        );
        autocannon.track(instance, {
            renderLatencyTable: true,
        });
    });
}
```

dnlup:
@mcollina @thekemkid What would be the best way to expose form-data submitting?

I was thinking exposing a separate option like:
```bash
autocannon --form '{ "encoding": "multipart/form-data", "body": { "image": { "type": "file", "path": "/path/to/file" }, "format": {"type": "text", "value": "jpegorelse"}}}'
```
that could be read also from file.

mcollina:
That works

tverilytt:
I wanted to run autocannon from the comnand-line with form file upload.
Ended up with something like this for GET / POST / DELETE of file.

'use strict'

const argv = require('minimist')(process.argv.slice(2));
const fs = require('fs');
const autocannon = require('autocannon');
const FormData = require('form-data');

let form;
if (argv.method === 'POST') {
[...]

PR #223, opened 2019-10-18
feat: add multipart form handling

dnlup:
- [x] implementation
- [x] tests
- [x] docs 

@mcollina This is a quick and hopefully not too dirty implementation of [what has been discussed about multipart forms](https://github.com/mcollina/autocannon/issues/160#issuecomment-537889310). If this approach is ok I can continue adding tests and docs.

Note:

this handles only multipart/form-data.


mcollina:
Would you mind:

1. remove package-lock, and add it to gitignore
2. add tests for this feature

Il giorno ven 18 ott 2019 alle 09:58 dnlup <notifications@github.com> ha
scritto:

dnlup:
@mcollina Sorry for the late response. I added tests, docs will follow too.

lib/run.js
```diff
@@ -93,13 +94,25 @@ function _run (opts, cb, tracker) {
...
+    }
+  }
 
   // copy over fields so that the client
   // performs the right HTTP requests
   url.pipelining = opts.pipelining
-  url.method = opts.method
-  url.body = opts.body
-  url.headers = opts.headers
+  url.method = opts.form ? 'POST' : opts.method
```
mcollina: I think we should honor the method option even if there is a form

test/runMultipart.test.js
```diff
@@ -0,0 +1,130 @@
...
+        })
+      })
+      await t.test('error', t => {
+        t.notEqual(null, err)
+        t.equal(c.message, err.message, `mismatching error message ${err.message}`)
+        t.end()
+      })
+    }))
+  }
+  await Promise.all(tests)
```
mcollina: There is no need to await them all, it’s done by tap automatically.

lib/multipart.js
```diff
@@ -0,0 +1,46 @@
...
+
+const { readFileSync } = require('fs')
+const FormData = require('form-data')
+
+function getFormData (string) {
+  try {
+    return JSON.parse(string)
+  } catch (error) {
+    try {
+      return require(string)
```
mcollina: Can you please use fe.readFileSync instead? require is relative to the folder of autocannon, not the current directory.

dnlup: Yes. This was my lazy approach to load both `json` and `js` files and I didn't think of that.

dnlup: would something like this also solve the issue?
```js
const path = require.resolve(string, { paths: [process.cwd()] })
return require(path)
```

mcollina: I would prefer the use of fs.readFile anyway, it shows intent.

dnlup: Ok

lib/multipart.js
```diff
@@ -0,0 +1,49 @@
...
+const FormData = require('form-data')
+
+function getFormData (string) {
+  try {
+    return JSON.parse(string)
+  } catch (error) {
+    try {
+      const path = resolve(string)
+      const data = readFileSync(path, 'utf8')
+      return JSON.parse(data)
```
dnlup: Here I am not allowing `.js` files anymore since it was being a little trickier than plain `.json`.

test/runMultipart.test.js
```diff
@@ -0,0 +1,117 @@
...
+      message: 'A \'type\' key with value \'text\' or \'file\' should be specified'
+    }
+  ]
+
+  for (const c of cases) {
+    t.test(c.name, async t => {
+      const [err] = await new Promise((resolve) => {
+        run({
+          url: 'http://localhost:' + server.address().port,
+          method: 'POST',
```
dnlup: Now that the `opts.method` in honored I am passing the method in the test calls.

mcollina: Ok!

lib/multipart.js
mcollina: that’s ok!

lib/run.js
dnlup: I have restored
```js
url.method = opts.method
```

but I am not sure that is what you meant anymore.

Maybe I could have assigned a default and let the user override it?

```js
if (opts.form) {
  opts.method = opts.method || 'POST'
}
```

mcollina: Yes, exactly.

dnlup:
Will do!

help.txt
```diff
@@ -47,7 +51,7 @@ Available options:
   -D/--reconnectRate NUM
         Some number of requests to make before resetting a connections connection to the
         server.
-  -n/--no-progress
+  -n/--no-progressallowed.includes(req.method)
```
mcollina: Is this change deliberate?

dnlup: No, sorry I'll fix it.

dnlup:
Sorry for the latest force pushes. I realized I committed with my work email address and GitHub was not recognizing me.

test/runMultipart.test.js
```diff
@@ -0,0 +1,141 @@
...
+      name: 'missing value in text type',
+      value: '{ "image": { "type": "text" }}',
+      message: 'Missing key \'value\' in form object for key \'image\''
+    },
+    {
+      name: 'missing value in text type',
+      value: '{ "image": { "type": "random" }}',
+      message: 'A \'type\' key with value \'text\' or \'file\' should be specified'
+    }
+  ]
```
mcollina: I think we should support an API where the form value is a JS object, and not just the json version of it

lib/run.js
```diff
@@ -93,13 +97,25 @@ function _run (opts, cb, tracker) {
...
 
   // copy over fields so that the client
   // performs the right HTTP requests
   url.pipelining = opts.pipelining
   url.method = opts.method
-  url.body = opts.body
-  url.headers = opts.headers
+  url.body = form ? form.getBuffer() : opts.body
+  url.headers = form ? {
+    ...Object.assign({}, opts.headers, form.getHeaders())
```
mcollina: there is no need for the object spread here

test/runMultipart.test.js
```diff
@@ -0,0 +1,160 @@
...
+      })
+      await t.test('error', t => {
+        t.notEqual(null, err)
+        t.equal(c.message, err.message, `mismatching error message ${err.message}`)
+        t.end()
+      })
+    })
+  }
+
+  t.end()
```
mcollina: I think this t.end is superflous

PR #224, opened 2019-10-29
Propagate counter on tick

catalin-me:
This PR adds the necessary logic to propagate the `counter` on the `tick` event. The new adjustments are also covered by integration tests.

### DIFF
--- a/lib/run.js
+++ b/lib/run.js
@@ -134,7 +134,7 @@ function _run (opts, cb, tracker) {
     totalCompletedRequests += counter
     requests.recordValue(counter)
     throughput.recordValue(bytes)
-    tracker.emit('tick', counter)
+    tracker.emit('tick', { counter, bytes })
     counter = 0
     bytes = 0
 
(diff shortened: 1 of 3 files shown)


catalin-me:
@mcollina Sure, I'll take care of it today. :)

catalin-me:
@mcollina I just updated the `README` as well. Please let me know your thoughts.

lib/run.js
```diff
@@ -134,9 +134,9 @@ function _run (opts, cb, tracker) {
     totalCompletedRequests += counter
     requests.recordValue(counter)
     throughput.recordValue(bytes)
+    tracker.emit('tick', counter)
```
thekemkid: What are your thoughts on making this line like this? 
```suggestion
    tracker.emit('tick', {counter, bytes})
```

If you think users would want access to the `count` per tick, would it be fair to think they'd want `bytes` per tick, too?

catalin-me: Great suggestion. I just updated the event. :)

PR #239, opened 2020-01-24
feat: add setupRequest per request for requests array

alex-ppg:
Hey all,

Since the `setupClient` appears to mutate the `Client` classs located in `httpClient.js`, I decided to add the ability to mutate each request independently within the `requests` array if it is provided. 

I thought about using the `setupClient` property at first, but it appears that each request in the `requests` array is passed to the `RequestIterator` and is never constructed as a `Client` presumably for performance reasons. For this purpose, I decided to name a new property called `setupRequest` that mutates the raw request object.

I have added corresponding tests, documentation and have followed the code style in autocannon. I hope this is a meaningful addition, as I myself did require this functionality in one of my projects where I tested the creation of entities and their subsequent mutation via a url param.

Please let me know if I should change anything in the PR to match your expectations.


mcollina:
Good work on the feature, the code seems ok for me.

alex-ppg:
Done, let me know if anything else is needed.

PR #242, opened 2020-01-31
Add expectBody option and related tests

doesdev:
This is probably a niche need driven by my paranoia. However, for certain scenarios it adds peace of mind knowing that the server sent what I expected it to back to the client. While this isn't necessarily the place of a benchmarking tool, it is helpful to me to have this for stress testing sorts of purposes.

I understand if it doesn't fit. That being said it's mostly an unobtrusive addition. The only thing it really imposes is the addition of `mismatches` to the API results. I suppose I could have fenced that off so that it only shows in the results if the option is enabled. Let me know if that would be desirable and I can do that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,6 +110,7 @@ Available options:
         Exclude error statistics (non 2xx http responses) from the final latency and bytes per second averages. default: false.
   -E/--expectBody EXPECTED
         Ensure the body matches this value. If enabled, mismatches count towards bailout.
+        Enabling this option will slow down the load testing.
   -v/--version
         Print the version number.
   -h/--help
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ Available options:
   -T/--title TITLE
         The title to place in the results for identification.
   -b/--body BODY
-        The body of the request. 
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -108,6 +108,8 @@ Available options:
         Server name for the SNI (Server Name Indication) TLS extension.
   -x/--excludeErrorStats
         Exclude error statistics (non 2xx http responses) from the final latency and bytes per second averages. default: false.
+  -E/--expectBody EXPECTED
+        Ensure the body matches this value. If enabled, mismatches count towards bailout.
```
mcollina:
```suggestion
        Ensure the body matches this value. If enabled, mismatches count towards bailout.
        Enabling this option will slow down the load testing.
```

doesdev: Added that warning to the README and help.txt

doesdev:
Awesome, thanks for the quick feedback. I've added the error and related test as well as that warning. Good call on `expectHeaders` and `expectStatusCode`. Those will be nice to have as well. Will work on those over the next couple days here. Thanks,

PR #244, opened 2020-02-15
Improve test coverage

leorossi:
Test coverage 100% on lines

@mcollina running tests reports a full 100% coverage on all metrics but some files are missing. This happened when I added `serial/autocannon.test.js` . If you run tests without that file you'll have 100% coverage an all files under `./lib`. Can you have a look?


mcollina:
Can you check CI? It failing.

I’m good with improving the situation anyway.

Issue #256, opened 2020-04-10
benchmark chunked encoding


yoshuawuyts:
A lot of traffic uses chunked encoding; this is common for streams that don't have a known size up front. This often takes a different route in HTTP parsers and encoders that isn't covered when benchmarking requests with a known length. It'd be fantastic if autocannon was able to issue repeated chunked requests.

Motivation for this is @dignifiedquire wanted to bench our chunked impl in https://github.com/http-rs/async-h1, and I didn't see an option for it in autocannon yet. Thanks!

mcollina:
That would be pretty interesting to do. A PR would be mostly welcomed.

evanxg852000:
I am currently catching up on my DB course. @mcollina Is it possible you guide me through? I would like to try a stab at this after the Easter break. That is from Tuesday 14th. Till then I am going to read the source code. Hopefully, I can post my questions here.

mcollina:
Sure, give a shout whenever you are ready!

nimit95:
I would also like to put in some effort too. Any way I can help?

mcollina:
I think this is a task for a single developer. Let's see how far we get along and then regroup, ok?

This repo has some issues - if you want to work on one feel free to pick!

evanxg852000:
@mcollina I have a WIP PR here to discuss around the API before going too far https://github.com/mcollina/autocannon/pull/258

PR #258, opened 2020-04-14
Support for http requests with chunked body [WIP]

evanxg852000:
Issue: https://github.com/mcollina/autocannon/issues/256
To: @mcollina @yoshuawuyts 

This is pull request is a WIP to add a chunked body HTTP request feature.
Currently, only the `lib/httpRequestBuilder.js`  has been modified to present the change and bootstrap discussion around the expected API.

In the current implementation, user is expected to pass `chunkSize` in the as option to the requestData. by default chunkSize is zero and represents the number of chars used to split the request body. That is if chunkSize is zero, the request is just a normal one. if the chunkSize is greater that zero the body string is split and converted to an array of buffer according to [HTTP spec](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Transfer-Encoding).
The other way of passing chunk to the query builder is to specify the body as an array of string or buffer.

Currently, only minimal work is done to start a discussion around the API. Once the API is agreed implementation should be updated or enhanced along with integration with the CLI args and tests should be written.

```js
const instance = autocannon({
  url: 'http://localhost:3000'
  chunkeSize: 5,
  body: 'the quick brown fox [<id>] jumps over the lazy dog.'
}, console.log)
```

```js
const instance = autocannon({
  url: 'http://localhost:3000'
  body: [ 'the quick brown ', 'fox [<id>] jumps ', 'over the lazy dog.']
}, console.log)
```

### DIFF
--- a/lib/httpRequestBuilder.js
+++ b/lib/httpRequestBuilder.js
@@ -20,10 +20,16 @@ function requestBuilder (defaults) {
 
   defaults = Object.assign(builderDefaults, defaults)
 
+  // cache chunks to avoid reprocessing
+  const chunkBodyCache = {}
   const chunkBody = (body, size) => {
     if (size === 0) {
       return body
     }
+    if (chunkBodyCache[body]) {
+      return chunkBodyCache[body]
+    }
+
     const chunks = []
     let found = body.search(/\[<id>\]/g)
     found = found === -1 ? false : found
@@ -39,6 +45,7 @@ function requestBuilder (defaults) {
       chunks.push(body.substr(start, size))
       start = start + size
     }
+    chunkBodyCache[body] = chunks
     return chunks
   }
 
@@ -47,10 +54,11 @@ function requestBuilder (defaults) {
       return [Buffer.from(chunks)]
     }
     const body = []
-    for (const chunk in chunks) {
+    for (let i = 0; i < chunks.length; ++i) {
+      const chunk = chunks[i]
       if (typeof chunk === 'string') {
         body.push(Buffer.from(`${chunk.length}\r\n`))
-        body.push(Buffer.from(chunk.join('\r\n')))
+        body.push(Buffer.concat([Buffer.from(chunk), Buffer.from('\r\n')]))
       } else if (Buffer.isBuffer(chunk)) {
         body.push(Buffer.from(`${chunk.length}\r\n`))
         body.push(Buffer.concat([chunk, Buffer.from('\r\n')]))
@@ -58,6 +66,7 @@ function requestBuilder (defaults) {
         throw new Error('body item must be either a string or a buffer')
       }
     }
+    body.push(Buffer.from('0\r\n')) // End of request marker
     return body
   }
 
@@ -134,7 +143,8 @@ function requestBuilder (defaults) {
     let req = Buffer.from(baseReq.join('\r\n') + '\r\n\r\n', 'utf8')
 
     if (bodyBuf && bodyBuf.length > 0) {
-      req = Buffer.concat([req, ...bodyBuf])
+      bodyBuf.unshift(req)
+      req = Buffer.concat(bodyBuf)
     }
 
     return req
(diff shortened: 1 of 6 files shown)


lib/httpRequestBuilder.js
```diff
@@ -13,12 +13,54 @@ function requestBuilder (defaults) {
...
+      if (found && (start + size) > found) {
+        chunks.push(body.substr(start, found - start))
+        chunks.push('[<id>]')
+        start = found + '[<id>]'.length
+        found = false
+        continue
+      }
+      chunks.push(body.substr(start, size))
+      start = start + size
+    }
```
mcollina, line 41: This is going to be _expensive_. Can we run it once and cache it?

lib/httpRequestBuilder.js
```diff
@@ -13,12 +13,54 @@ function requestBuilder (defaults) {
...
+    return chunks
+  }
+
+  const makeChunkedBody = (chunks) => {
+    if (typeof chunks === 'string') {
+      return [Buffer.from(chunks)]
+    }
+    const body = []
+    for (const chunk in chunks) {
```
mcollina, line 50: can you use a for(;;) loop?

lib/httpRequestBuilder.js
```diff
@@ -85,7 +134,7 @@ function requestBuilder (defaults) {
     let req = Buffer.from(baseReq.join('\r\n') + '\r\n\r\n', 'utf8')
 
     if (bodyBuf && bodyBuf.length > 0) {
-      req = Buffer.concat([req, bodyBuf])
+      req = Buffer.concat([req, ...bodyBuf])
```
mcollina, line 137: `...` is kind of expensive, we should avoid it.

lib/httpRequestBuilder.js
evanxg852000: Thanks for the feedback, since you seem to be happy with the direction, I will now keep on writing it and improving.

lib/httpRequestBuilder.js
evanxg852000: Thanks for teaching me this. an old `unshift` should have been fine but wanted to be cool 😄

evanxg852000:
@mcollina I have integrated my change. however, I get timeout report on all emited request when I try `./autocannon.js -b "hello world" -k 2  -m GET http://localhost:3000`
the params are set correctly and I can see the test server is returning the correct response.
The above command generates the following request"
```
GET / HTTP/1.1
Host: localhost:3000
Transfer-Encoding: chunked

2
he
2
ll
2
o
2
wo
2
rl
1
d
0

```
I will keep debugging but let me know if you have a tip. Thanks

lib/httpRequestBuilder.js
```diff
@@ -13,12 +13,63 @@ function requestBuilder (defaults) {
...
+      }
+      chunks.push(body.substr(start, size))
+      start = start + size
+    }
+    chunkBodyCache[body] = chunks
+    return chunks
+  }
+
+  const makeChunkedBody = (chunks) => {
+    if (typeof chunks === 'string') {
```
mcollina, line 53: I think this could be a buffer as well.

lib/httpRequestBuilder.js
```diff
@@ -13,12 +13,63 @@ function requestBuilder (defaults) {
...
     hostname: 'localhost',
+    chunkSize: 0,
     setupRequest: reqData => reqData,
     port: 80
   }
 
   defaults = Object.assign(builderDefaults, defaults)
 
+  // cache chunks to avoid reprocessing
+  const chunkBodyCache = {}
```
mcollina, line 24: This might leak memory if the body is changed. I would recommend using an lru cache, such as the one provided in the tiny-lru package.

rundlof:
Great tool! I would be very interested to know if you are planning to finish support for chunked body?

mcollina:
@rundlof if you'd like to take over this task and send a fresh PR, go for it!

Bolek719:
Duplicate of #

Bolek719:
@Bolek719

mcollina:
Closing for no activity

PR #261, opened 2020-06-05
feat: Reponse handling and request context

feugy:
### What's in there?

We would like, when using autocannon's node API, to have more control over requests.
In particular, we need to get the response of the previous request, extract some data out of it, and use it for the next request.

#### The proposed approach

It is already possible to customize sent request with `setupRequest(req) { return req }` function.
The idea is:
 1. to add an extra parameter: a context object that would last until autocannon loops over available requests.
 2. to support an `onResponse(status, body, context) {}` function, that can massage server response and enrich context passed to the next request.

```js
  // from samples/request-context.js

  autocannon({
    url: url,
    requests: [
      {
        // let's create a new user
        method: 'POST',
        path: '/users',
        body: JSON.stringify({ firstName: 'Jane', id: 10 }),
        onResponse: (status, body, context) => {
          if (status === 200) {
            context.user = JSON.parse(body)
          } // on error, you may abort the benchmark
        }
      },
      {
        // now we'll give them a last name
        method: 'PUT',
        setupRequest: (req, context) => ({
          ...req,
          path: `/user/${context.user.id}`,
          body: JSON.stringify({
            ...context.user,
            lastName: 'Doe'
          })
        })
      }
    ]
  })
```

#### Why not using `response` event from autocannon?

This event is fired when autocannon received the response header. It does not have access to the body.

Besides, autocannon generates all request bodies once for all, before it enters the loop, which makes customization very difficult (we would need to use client's `setBody()` function.

#### Drawbacks of this approach

`setupRequest()` function is now called right before issuing request, instead of once for all, before starting benchmark.

These is a breaking change in autocannon's current behaviour. However, I'll argue it makes much sense, for such dynamic function, to behaves this way. 
I'm totally open to other approaches! Please challenge this one.

### Notes to reviewers

Because some request may care about their response, and some other won't, `HttpClient`will trigger next request accordingly.
- when current request has no `onResponse`: the `reponse` event is emitted when response headers are received, and next request is triggered (current behaviour)
- when current request has an `onResponse` function: the `response` event is emitted when the body is received, and the next request is triggered.

`RequestIterator` offers much more possibilities than `HttpClient`needs. Applying changes and keeping a consistent API was tricky: its API is both flexible and pervasive:
- `move()` was the only way to substitute id in the request buffer. But it was also moving to next request. 
- `setRequests()` was computing all request buffers at once.

For the reasons explained above, the new approach is simpler and more consistent:
1. `nextRequest()` is the only whay to move the iterator
1. `rebuildRequest()` is handling id substitution, ensuring that request buffer is always ready to be sent.

### DIFF
--- a/lib/httpClient.js
+++ b/lib/httpClient.js
@@ -86,29 +86,43 @@ function Client (opts) {
     this.resData[this.cer].headers = opts
   }
 
+  const emitAndMove = () => {
+    const resp = this.resData[this.cer]
+    this.emit('response', resp.headers.statusCode, resp.bytes, resp.duration)
+    resp.bytes = 0
+    this.cer = this.cer === opts.pipelining - 1 ? 0 : this.cer++
+    this.requestIterator.nextRequest()
+    this._doRequest(this.cer)
+  }
+
   this.parser[HTTPParser.kOnBody] = (body, start, len) => {
     this.emit('body', body)
+    const bodyString = '' + body.slice(start, start + len)
 
     if (this.opts.expectBody) {
-      const bodyString = '' + body.slice(start, start + len)
       if (this.opts.expectBody !== bodyString) {
         return this.emit('mismatch', bodyString)
       }
+    }
+
+    const resp = this.resData[this.cer]
+    if (this.requestIterator.expectsBody()) {
+      this.requestIterator.recordBody(resp.headers.statusCode, bodyString)
+      emitAndMove()
     }
   }
 
   this.parser[HTTPParser.kOnMessageComplete] = () => {
     const end = process.hrtime(this.resData[this.cer].startTime)
-    const responseTime = end[0] * 1e3 + end[1] / 1e6
-    this.emit('response', this.resData[this.cer].headers.statusCode, this.resData[this.cer].bytes, responseTime)
-    this.resData[this.cer].bytes = 0
+    this.resData[this.cer].duration = end[0] * 1e3 + end[1] / 1e6
 
     if (!this.destroyed && this.reconnectRate && this.reqsMade % this.reconnectRate === 0) {
       return this._resetConnection()
     }
 
-    this.cer = this.cer === opts.pipelining - 1 ? 0 : this.cer++
-    this._doRequest(this.cer)
+    if (!this.requestIterator.expectsBody()) {
+      emitAndMove()
+    }
   }
 
   this._connect()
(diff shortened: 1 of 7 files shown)


lib/httpClient.js
```diff
@@ -86,29 +86,43 @@ function Client (opts) {
     this.resData[this.cer].headers = opts
   }
 
+  const emitAndMove = () => {
+    const resp = this.resData[this.cer]
+    this.emit('response', resp.headers.statusCode, resp.bytes, resp.duration)
+    resp.bytes = 0
+    this.cer = this.cer === opts.pipelining - 1 ? 0 : this.cer++
```
mcollina, line 93: Can you please avoid getting the value from `opts.pipelining` and set `pipeling` to a variable on top?

lib/requestIterator.js
```diff
@@ -19,7 +20,11 @@ inherits(RequestIterator, Object)
 RequestIterator.prototype.nextRequest = function () {
   ++this.currentRequestIndex
   this.currentRequestIndex = this.currentRequestIndex < this.requests.length ? this.currentRequestIndex : 0
+  if (this.currentRequestIndex === 0) {
+    this.context = {}
+  }
```
mcollina, line 25: Can you please avoid this if and initialize this in the constructor?

feugy: it is done as well, but the point is to clear the context when restarting the loop.

mcollina: Ah ok, can you add a comment then?

feugy:
Unfortunately, yes, it reduce the throughput quite a lot:
> 536k requests in 11.04s, 59.5 MB read (master)
> 320k requests in 10.04s, 35.5 MB read (this PR)

Let me see if I can change that.

feugy:
Ok, it's there @mcollina.

I've make sure that all request which don't have `setupRequest()` function are built prior to the test run.
Only the ones with `setupRequest()`will be built on the fly. 

This ensures that CLI usage and simple programmatic usage will be as efficient as it used to be.

lib/requestIterator.js
mcollina: You can include this if in the previous one. 

```js
if (this.currentRequestIndex === this.requests.length) {
  this.currentRequestIndex = 0
  // when looping over available request, clear context for a fresh start
  this.context = {}
}
```

mcollina:
I confirm this has no performance impact for the nominal case.

mcollina:
Would you mind to add docs?

PR #274, opened 2020-08-03
feat: add a correct coordinated omission option

alexvictoor:
Hello!
This is a small PR to add a `correct coordinated omission` feature to autocannon. 
It works with either connectionRate or overallRate options. 
I have left it deactivated by default since the impact on the displayed latencies can be quite significant...
Let me know if you need me to do any change

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,6 +91,8 @@ Available options:
         connection rate will take precedence if both are set.
         NOTE: if using rate limiting and a very large rate is entered which cannot be met,
               Autocannon will do as many requests as possible per second.
+  -C/--correctCoordinatedOmission
+        Correct coordinated omission issue. Can be used in conjunction with 'connectionRate' or 'overallRate'.
   -D/--reconnectRate NUM
         Some number of requests to make before resetting a connections connection to the
         server.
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -91,6 +91,8 @@ Available options:
         connection rate will take precedence if both are set.
         NOTE: if using rate limiting and a very large rate is entered which cannot be met,
               Autocannon will do as many requests as possible per second.
+  -C/--correctCoordinatedOmission
```
mcollina, line 94: Can you add some docs to explain why this is needed? Also, it seems something we might want to enable by default? Why somebody would not want this?

lib/run.js
```diff
@@ -258,12 +258,18 @@ function _run (opts, cb, tracker) {
...
     }
 
-    function onResponse (statusCode, resBytes, responseTime) {
+    function onResponse (statusCode, resBytes, responseTime, rate) {
       tracker.emit('response', this, statusCode, resBytes, responseTime)
       const codeIndex = Math.floor(parseInt(statusCode) / 100) - 1
       statusCodes[codeIndex] += 1
-      // only recordValue 2xx latencies
-      if (codeIndex === 1 || includeErrorStats) latencies.recordValue(responseTime)
+      // only recordValue 2xx latenciesxz
```
mcollina, line 265: why xz?

alexvictoor: Oups... I guess this is a typo

README.md
alexvictoor: Well you are right it could be enable by default, like in wrk2. The downside would be for users that do benchmarks on a regular basis and track the performance trends. Do not know if this would impact many users though...  
If It is enable by default would you like to be able to deactivate it from the command line or would you like to get in the output corrected and uncorrected results (like in wrk2)?

mcollina: I would recommend to deactivate from the command line. I would also add some reference on the problem as a lot of people would be unfamiliar with it.

alexvictoor:
I have tried to take in account your feedback. Not that easy to explain CO... Let me know if there is anything else I can change/improve

README.md
```diff
@@ -90,7 +90,10 @@ Available options:
         The max number of requests to make per second from an all connections.
         connection rate will take precedence if both are set.
         NOTE: if using rate limiting and a very large rate is entered which cannot be met,
-              Autocannon will do as many requests as possible per second.
+              Autocannon will do as many requests as possible per second. Also latency data will be corrected in order to compensate the effects of coordinated omission issue. If you are not familiar with the coordinated omission issue, you should probably read [this article](http://highscalability.com/blog/2015/10/5/your-load-generator-is-probably-lying-to-you-take-the-red-pi.html) or watch this [Gil Tene's talk](https://www.youtube.com/watch?v=lJ8ydIuPFeU) on the topic.
+  -C/--ignoreCoordinatedOmission
+        Ignore coordinated omission issue when requests should be sent at a fixed rate using 'connectionRate' or 'overallRate'. 
+      NOTE: it is not recommanded to enable this option. When the request rate cannot be met because the server is too slow, many request latencies might be missing and Autocannon might report a misleading latency distribution.
```
mcollina, line 96: Can you sync this text in https://github.com/mcollina/autocannon/blob/master/help.txt?

alexvictoor:
That would be great!

PR #276, opened 2020-08-15
fix: Report (full) response body correctly (#275)

### LINKED ISSUE #275: Response body is not handled correctly
I've found the following issue in the response handling:

## The issue

When the response data is sufficiently large (not sure what the threshold is), `onResponse` is called multiple times, each time with a partial body. This makes it really hard to e.g. validate the response body, like checking whether it is valid JSON and matches a certain structure.

The `expectedBody` feature is also broken because it's checked multiple times, again against partial body data, instead of the full response body.

## The reason

This is because `Client` checks `expectedBody` and calls `onResponse` every time `this.parser[HTTPParser.kOnBody]` fires. However, this can happen multiple times per response, and the handler will be passed a partial body every time.

## The fix

This can be fixed by

1. recording the full body in `resData[this.cer]` (e.g. via a new `body` field), and
2. only checking `expectedBody` and calling `onResponse` (via `this.requestIterator.recordBody()`) once in the `HTTPParser.kOnMessageComplete` handler.

Jannis:
This fixes #275 by implementing the fix described in the issue.

### DIFF
--- a/lib/httpClient.js
+++ b/lib/httpClient.js
@@ -49,6 +49,7 @@ function Client (opts) {
   for (let i = 0; i < pipelining; i++) {
     this.resData[i] = {
       bytes: 0,
+      body: '',
       headers: {},
       startTime: [0, 0]
     }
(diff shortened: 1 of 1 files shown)


Jannis:
Yup, I can try!

Jannis:
@mcollina I'm actually not sure how to simulate a response that would lead to `HTTPParser.kOnBody` being called twice. I've been able to test this fix against a real server though.

mcollina:
Maybe you can use a mock.

blanchma:
Hi guys, we have being following this issue because we had it too. Is this going to merge?

mcollina:
Not in this current form as it is missing a test. Would you like to contribute one?

salmanm:
@mcollina Let's close this.

Issue #233, opened 2019-11-29
SNI doesn't work without config


LinusU:
Trying to benchmark against a service running behind API Gateway + Lambda on Amazon resulted in the connection not working (#232). Turns out that the SNI negotiation didn't work, and passing `--servername myhost.com` (not actual hostname) fixed it.

Since it wasn't my personal server I'm hesitant to give out the URL, but I think it should be easy to test against any server that requires SNI in order to negotiate HTTPS.

failing example:

```sh
autocannon -c 1 -a 1 https://foobar.myhost.com/
```

working example:

```sh
autocannon -c 1 -a 1 https://foobar.myhost.com/ --servername foobar.myhost.com
```

silverwind:
Also ran into this when ranning against a SNI-only service. I think this module should by default send SNI servername based on the hostname in the URL to make this more intuitive and closer to actual client requests.

mcollina:
Would you like to send a Pull Request to address this issue? Remember to add unit tests.

silverwind:
Yes, I'll look into it.

PR #277, opened 2020-08-20
Enable SNI servername by default

silverwind:
Default to sending the SNI servername when the `servername` option is not present or falsy. This behaviour matches browsers as well as popular tools like `curl` or `openssl s_client`.

Fixes: https://github.com/mcollina/autocannon/issues/233

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,7 +109,7 @@ Available options:
   -f/--forever
         Run the benchmark forever. Efficiently restarts the benchmark on completion. default: false.
   -s/--servername
-        Server name for the SNI (Server Name Indication) TLS extension.
+        Server name for the SNI (Server Name Indication) TLS extension. Defaults to the hostname of the URL when it is not an IP address.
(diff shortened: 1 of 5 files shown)


silverwind:
Looks like some extra logic is needed to filter out IP adresses.

````
(node:1530) [DEP0123] DeprecationWarning: Setting the TLS ServerName to an IP address is not permitted by RFC 6066. This will be ignored in a future version.
````

Also, I'm wondering if `''` vs `undefined` servername makes a difference. If the empty string results in the extension being disabled in node, it'd provide a convenient way to opt-out of SNI.

silverwind:
Fixed that (imho questionable) deprecation by not inferring servername from IP addresses in the hostname. It'll still be possible to specify them explicitely as I don't want to prevent users doing that (IP adresses itself are not per-se invalid servernames).

Also fixed a bug related to variable closure in `startTlsServer` where one test case would see the `servername` of another.

PR #281, opened 2020-09-09
feat: restart request iterator when setupRequest returns falsey

feugy:
`setupRequest()` is very handy to perfom dynamic scenarios.
However, users do not necessarily realize they have to return a request object. If they don't, it'll crash the whole benchmark.

Besides a very common need is to skip the rest of the scenario.

This feature fixes both, by reset request iterator based on `setupRequest()`.


mcollina (approved):
lgtm

PR #282, opened 2020-09-10
feat: report many pipeline resets

feugy:
This is a follow up of #281 
Now, request iterator resets when some of its `setupRequest()` returning falsey value (previously it the pipeline was simply crashing).

But this may actually obfuscate bugs of the tested service: often `setupRequest()` fails as the previous response did not contained expected data.

This PR simply displays how many times the pipeline reseted, so users could decide whether it is intended or not.

```shell
┌─────────┬────────┬────────┬────────┬────────┬───────────┬───────────┬────────────┐
│ Stat    │ 2.5%   │ 50%    │ 97.5%  │ 99%    │ Avg       │ Stdev     │ Max        │
├─────────┼────────┼────────┼────────┼────────┼───────────┼───────────┼────────────┤
│ Latency │ 121 ms │ 624 ms │ 890 ms │ 914 ms │ 541.06 ms │ 237.32 ms │ 1002.16 ms │
└─────────┴────────┴────────┴────────┴────────┴───────────┴───────────┴────────────┘
┌───────────┬────────┬────────┬────────┬────────┬────────┬─────────┬────────┐
│ Stat      │ 1%     │ 2.5%   │ 50%    │ 97.5%  │ Avg    │ Stdev   │ Min    │
├───────────┼────────┼────────┼────────┼────────┼────────┼─────────┼────────┤
│ Req/Sec   │ 218    │ 218    │ 372    │ 479    │ 356.5  │ 70.29   │ 218    │
├───────────┼────────┼────────┼────────┼────────┼────────┼─────────┼────────┤
│ Bytes/Sec │ 190 kB │ 190 kB │ 372 kB │ 465 kB │ 369 kB │ 67.3 kB │ 190 kB │
└───────────┴────────┴────────┴────────┴────────┴────────┴─────────┴────────┘

Req/Bytes counts sampled once per second.

10k requests in 28.09s, 10.3 MB read
18 errors (0 timeouts)
request pipeline was resetted 4 times

```

_Note:_ I know `test/progressTracker.test.stub.js` isn't part of the test suite, but as I've changed it to troubleshoot the feature, I though it would be fine to keep a new test in it.
The feature is properly covered in `est/httpClient.test.js` and `test/run.test.js` test files.

_Edit:_ this PR actually unveiled some flaws in the previous implementation.
Reseting request pipeline the way I did is wrong, as it may leave `currentRequest.requestBuffer` to `null`.

This new version fixes the flaw, and also handle a specific case where the very first `setupRequest()` fails. In this situation, we can not reset pipeline (or it'll never end), and we must exit on an error.

### DIFF
--- a/lib/progressTracker.js
+++ b/lib/progressTracker.js
@@ -132,7 +132,7 @@ function track (instance, opts) {
       logToStream(`${format(result.mismatches)} requests with mismatched body`)
     }
     if (result.resets) {
-      logToStream(`request pipeline was reseted ${format(result.resets)} ${result.resets === 1 ? 'time' : 'times'}`)
+      logToStream(`request pipeline was resetted ${format(result.resets)} ${result.resets === 1 ? 'time' : 'times'}`)
     }
   })
 
--- a/lib/progressTracker.js
+++ b/lib/progressTracker.js
@@ -131,6 +131,9 @@ function track (instance, opts) {
     if (result.mismatches) {
       logToStream(`${format(result.mismatches)} requests with mismatched body`)
     }
+    if (result.resets) {
+      logToStream(`request pipeline was reseted ${format(result.resets)} ${result.resets === 1 ? 'time' : 'times'}`)
+    }
   })
 
   function logToStream (msg) {
--- a/README.md
+++ b/README.md
@@ -342,6 +342,7 @@ The results object emitted by `done` and passed to the `autocannon()` callback h
 * `connections`: The amount of connections used (value of `opts.connections`).
 * `pipelining`: The number of pipelined requests used per connection (value of `opts.pipelining`).
 * `non2xx`: The number of non-2xx response status codes received.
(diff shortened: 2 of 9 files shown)


lib/progressTracker.js
```diff
@@ -131,6 +131,9 @@ function track (instance, opts) {
     if (result.mismatches) {
       logToStream(`${format(result.mismatches)} requests with mismatched body`)
     }
+    if (result.resets) {
+      logToStream(`request pipeline was reseted ${format(result.resets)} ${result.resets === 1 ? 'time' : 'times'}`)
```
mcollina, line 135:
```suggestion
      logToStream(`request pipeline was resetted ${format(result.resets)} ${result.resets === 1 ? 'time' : 'times'}`)
```

mcollina (approved):
lgtm

Issue #272, opened 2020-07-31
Use webassembly hdrhistogram


mcollina:
See https://github.com/HdrHistogram/HdrHistogramJS#boosting-performances-with-webassembly

dnlup:
I think this one can be closed, right?

PR #283, opened 2020-09-14
Enable WASM for HdrHistogramJS

alexvictoor:
Hello
This PR is related to [issue 272](https://github.com/mcollina/autocannon/issues/272)
The two things that are very important when using wasm histograms are to load the binary prior to instantiation and to free the memory when the histograms are not needed anymore.
Let me know if you prefer to enable wasm histograms only when a config switch is on.

### DIFF
--- a/lib/run.js
+++ b/lib/run.js
@@ -36,7 +36,10 @@ function _run (opts, cb, tracker) {
     return tracker
   }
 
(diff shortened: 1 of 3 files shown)


mcollina (approved):
lgtm

PR #285, opened 2020-09-29
feat: support HAR files as input

### LINKED ISSUE #185: Session recording / HAR support
As titled!

feugy:
#### What's in there?

- [x] support for [HAR](https://w3c.github.io/web-performance/specs/HAR/Overview.html) files as request input
- [x] closes #185

This PR allow CLI users to speciffy `--har`  options with the path to an `.har`  file.
Requests from the har will be fired against the specify server.

Please note that Autocannon will ignore HAR requests that would not match the domain of your URL.
This is because each httpClient handles the connection to a single domain. It is not possible for a client to switch to a different one, based on current request. 

#### How to test?

You can use any of the test fixtures:
```shell
node autocannon --har ./test/fixtures/httpbin-get.har -d 3 https://httpbin.org
``` 

You can also record an .har from your favorite browser's dev tools. 
For example, with you devtools opened, go the [node.js reference](https://nodejs.org/api/url.html#url_legacy_url_api). A couple of requests will appear on network tab. 
Export them as an HAR file, and try playing it again:

```shell
node autocannon --har test.har -d 3 -j https://nodejs.org
``` 

Please note that some request may "fail" with "3xx" statuses.

Finally, it is possible to specify `har`  option when using the programmatic API, as demonstrated in the `tests/run.test.js` file.

#### Notes to reviewers

- HAR files are plain JSON files, and there is no need for a specific parser, in particular because we're using a tiny part of the file (we ignore pages, responses, timing...)

- `parseHAR.js`  file is exposing functions to read and parse files in a synchronous or asynchronous fashion. However, it made little sense to use the _asynchronous_ flavor: HAR parsing occurs only once, before starting benchmark, and that part of Autocannon code if synchronous.
Changing it to asynchronous would be a major change for little-to-no benefit (in my opinion).

- there could be more checks on the HAR content (I assumed that requests will contain the proper keys), and I'm happy to add more if you think it worth it.


lib/parseHAR.js
```diff
@@ -0,0 +1,55 @@
...
+  }
+}
+
+function parseHAR (file, cb) {
+  fs.readFile(file, 'utf8', function (err, data) {
+    if (err) {
+      return cb(err)
+    }
+    try {
+      cb(null, parse(data))
```
mcollina, line 43: note that this will call the callback twice if cb throws. Please move the call to cb outside of the try-catch.

test/run.test.js
```diff
@@ -697,3 +697,87 @@ test('should count resets', t => {
     t.end()
   })
 })
+
+test('should parse HAR file', (t) => {
+  t.plan(6)
+
+  run({
+    url: 'https://httpbin.org',
```
mcollina, line 705: Are we calling httpbin directly? Maybe use an internal server?

README.md
```diff
@@ -78,6 +78,10 @@ Available options:
         The body of the request. See '-b/body' for more details.
   -H/--headers K=V
         The request headers.
+  --har FILE
+        When provided, Autocannon will use requests from the HAR file. 
+        CAUTION: you have to to specify one (or more) domain using URL option: only the HAR requests to the same domains will be considered.
```
mcollina, line 83: Does this throw? Is the domain overridden?

README.md
```diff
@@ -246,12 +250,13 @@ Start autocannon against the given target.
...
+       * `headers`: When present, will override `opts.headers`. _OPTIONAL_
+       * `method`: When present, will override `opts.method`. _OPTIONAL_
+       * `path`: When present, will override `opts.path`. _OPTIONAL_
+       * `setupRequest`: A `Function` you may provide to mutate the raw `request` object, e.g. `request.method = 'GET'`. It takes `request` (Object) and `context` (Object) parameters, and must return the modified request. When it returns a falsey value, autocannon will restart from first request. _OPTIONAL_
+       * `onResponse`: A `Function` you may provide to process the received response. It takes `status` (Number), `body` (String) and `context` (Object) parameters. _OPTIONAL_
+    * `har`: the `String` location of an [HAR](https://w3c.github.io/web-performance/specs/HAR/Overview.html) file. Autocannon will extract request from this file, keeping only the ones targeting the same domains as the given urls. Requests method, path, query parameters, headers and body will be set from the HAR file. _NOTE_: this will take over `requests` property. You can add extra headers with `headers` property, but `method`, `form` and `body` will be ignored. _OPTIONAL_
```
mcollina, line 259: I think it's a path on disk. It would be also ok to support it as a Javascript object.

mcollina (changes requested):

README.md
feugy: Neither of the two.
if you specify URL https://A.com, requests from the HAR file to https://B.com will simply be skipped.
Please note that we support multiple domains:
- assuming HAR contains requests to A, B and C,
- assuming you specify A and B urls,
- the first connection will fire requests to A, 
- the next connection will fire requests to B,
- requests to C will be ignored.

Overriding would not make sense, as you're likely to issue a request to an unsupported path or with unsupported parameters.

We can throw as well, but supporting the case above will be harder.

README.md
feugy: Correct, it's the path on disk. I'll change the comment.
About JSON content: do you want to support both, or only JSON?

test/run.test.js
feugy: Yes we are.

The difficulty with using an interval server is to change the har file.
We'll need to change the urls in file to match the internal server, which port is dynamic. 
It can be done by copying the fixture file, altering the copy, use the copy during test, and remove it.

I though it was ok to use httpbin, given it's meant for testing. If you prefer, I can change that in favor of the solution above.

mcollina: Please do the above. Touching the network make the test not pass in case of an outage.

README.md
mcollina: I would just support a JS object in HAR format and leave the parsing for the CLI.

README.md
mcollina: I don't really understand this behavior and it looks like it's quite confusing. Let's just throw for now and then we leave this to further iterations? (Open an issue after landing)

feugy: It would be very hard to use.
Tools that generate HAR like browsers usually wraps requests from many domains.
People will have to alter their files accordingly to trim out requests they don't want.

Just to be sure about your intention:
Given user can pass several urls from the CLI, do you want to throw an error if the HAR file contains a requests that does not target one of these urls?

mcollina: > It would be very hard to use. Tools that generate HAR like browsers usually wraps requests from many domains.
People will have to alter their files accordingly to trim out requests they don't want.

I understand now! Maybe we can print a warning about the domains that are ignored?

mcollina (approved):
lgtm

mcollina:
Can you please update the PR text saying that it implements #185?

mcollina (approved):
lgtm

PR #287, opened 2020-10-16
fix: Record bodies in the RequestIterator even if the body is empty.

ShogunPanda:
We have noticed in our API use of autocannon that `onResponse` callback are not invoked if the server sends no body.
This PR fixes the issue.

### DIFF
--- a/test/run.test.js
+++ b/test/run.test.js
@@ -698,6 +698,35 @@ test('should count resets', t => {
   })
 })
 
+test('should get onResponse callback invoked even when there is no body', t => {
+  t.plan(4)
+  const server = helper.startServer({ responses: [{ statusCode: 200, body: 'ok' }, { statusCode: 204 }] })
+
+  run({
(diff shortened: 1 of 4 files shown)


test/run.test.js
```diff
@@ -698,6 +698,35 @@ test('should count resets', t => {
   })
 })
 
+test('should get onResponse callback invoked even when there is no body', t => {
```
mcollina, line 701: please use async/await instead of `then`.

mcollina (changes requested):
There seem to be a failing test on Node v12 on Windows :/. I'm not really sure how would you like to address that.

test/run.test.js
ShogunPanda: Done

ShogunPanda:
It seems like the build is timing out. I've updated the tap's timeout to see if it fixes it.

mcollina (approved):
lgtm

PR #288, opened 2020-10-17
Limitations

mcollina:
Explain a bit more the implied limitations of autocannon.

### DIFF
--- a/README.md
+++ b/README.md
@@ -426,6 +426,16 @@ Nevertheless, it uses significant more CPU than other tools that compiles to a b
 Autocannon can saturate the CPU, e.g. the autocannon process reaches 100%: in those cases
 we recommend to use `wrk`.
 
+As an example, let's consider a run with 1000 connections on a server
+with 4 cores with hyperhtreading:
+
+* `wrk` uses 2 threads (by default) and an auxiliary one to collect the
+  metrics with a total load of the CPU of 20% + 20% + 40%.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -416,6 +416,19 @@ function handleResults(result) {
...
 
+<a name="limitations"></a>
+## Limitations
+
+Autocannon is written in JavaScript for the Node.js runtime and it is CPU-bound.
+We have verified that it yields comparable results with `wrk` when benchmarking Node.js
+applicatiions using the `http` module.
+Nevertheless, it uses significant more CPU than other tools that compiles to a binary such as `wrk`.
+Autocannon can saturate the CPU, e.g. the autocannon process reaches 100%: in those cases
+we recommend to use `wrk`.
```
jasnell, line 427: In saturation cases, I've also run autocannon from an adjacent machine. You've got to account for network latency but once you do the perf analysis still works.

mcollina: Unfortunately it can still happen that autocannon saturates the CPU.

Let's make an example with 1000 connections on my dedicated server with 4 cores with hyperthreading:

* wrk uses 2 threads + an auxiliary one to collect data: 20% + 20% + 40% cpu load.
* autocannon uses one thread at 80% cpu load.

Both saturates a Node.js process at around 41k req/sec, however wrk could potentially create more load due to its multi threading nature (because the server has plenty of CPU power available).

(I will include this example)

---

The solution is to use piscina and run workers!

PR #290, opened 2020-10-21
feat: introduce support to use worker threads

salmanm:
This PR adds support to use multiple worker threads to fire requests. The flag `--numWorkers`/`-w` can be passed as command-line argument with a number that will be used to fire up the workers.


lib/progressTracker.js
```diff
@@ -84,56 +81,13 @@ function track (instance, opts) {
   }
 
   instance.on('done', (result) => {
+    if (opts.useWorkers) return
```
mcollina, line 84: Let's not add the event listener if `useWorker` is set.

lib/worker.js
```diff
@@ -0,0 +1,35 @@
...
+    tracker.stop()
+  })
+}
+
+if (!isMainThread) {
+  const { opts } = workerData
+
+  runTracker(opts, (result) => {
+    parentPort.postMessage(result)
+  })
```
mcollina, line 34: Unfortunately I fear this won't work. I don't think it's possible to combine multiple histograms into one. We need to put all the data in the same histogram, i.e. the workers should send the samples in.

lib/worker.js
```diff
@@ -0,0 +1,35 @@
...
+const track = require('./progressTracker')
+const run = require('./run')
+
+function runTracker (argv, ondone) {
+  const tracker = run(argv)
+
+  tracker.on('done', (result) => {
+    if (ondone) ondone(result)
+    if (argv.json) {
+      console.log(JSON.stringify(result))
```
mcollina, line 11: This should not console.log.

lib/worker.js
```diff
@@ -0,0 +1,35 @@
+const { isMainThread, parentPort, workerData } = require('worker_threads')
```
mcollina, line 1: use strict is missing.

mcollina (changes requested):

autocannon.js
```diff
@@ -243,6 +243,9 @@ function createChannel (onport) {
 function runTracker (argv, ondone) {
   const tracker = run(argv)
 
+  // While using workers, tracking is handled in manager.js
+  if (argv.useWorkers) return
```
mcollina, line 247: I would prefer we move `runTracker` to its own file, and call it or `manager.js` depending if we are using workers or not.

lib/aggregateResult.js
```diff
@@ -0,0 +1,69 @@
+const hdr = require('hdr-histogram-js')
```
mcollina, line 1: use strict is missing

lib/manager.js
```diff
@@ -0,0 +1,71 @@
...
+const numWorkers = Math.max(Math.floor(os.cpus().length * 0.75), 1)
+
+function initWorkers (opts, cb) {
+  const workers = []
+  const results = []
+
+  function startAll () {
+    spinner.start()
+
+    workers.forEach(w => {
```
mcollina, line 18: use for..of

lib/manager.js
```diff
@@ -0,0 +1,71 @@
...
+spinner.setSpinnerString('|/-\\')
+
+const numWorkers = Math.max(Math.floor(os.cpus().length * 0.75), 1)
+
+function initWorkers (opts, cb) {
+  const workers = []
+  const results = []
+
+  function startAll () {
+    spinner.start()
```
mcollina, line 16: Please check that the spinner does not scramble output if using a non-TTY, e.g. redirect the output to a file.

lib/manager.js
```diff
@@ -0,0 +1,71 @@
...
+          spinner.stop(true)
+          cb(results)
+        }
+      }
+    })
+
+    worker.on('error', (err) => {
+      console.log('Worker error:', err)
+      stopAll()
+      process.exit(1)
```
mcollina, line 58: I prefer we call the callback with the error instead.

lib/run.js
```diff
@@ -4,18 +4,44 @@ const EE = require('events').EventEmitter
...
+  // the code below this `if` just renders the results table...
+  // if the user doesn't want to render the table, we can just return early
+  if (!isMainThread || opts.renderResultsTable === false) return
+
+  printResult(result, opts)
+}
 
 function run (opts, cb) {
-  return _run(opts, cb)
+  if (opts.useWorkers && isMainThread) {
```
mcollina, line 32: Can we avoid using `isMainThread` and pass around options instead?

lib/worker.js
```diff
@@ -0,0 +1,47 @@
...
+  })
+
+  tracker.on('error', (err) => {
+    if (err) {
+      throw err
+    }
+  })
+
+  // if not rendering json, or if std isn't a tty, track progress
+  if (!argv.json || !process.stdout.isTTY) track(tracker, argv)
```
mcollina, line 25: I don't understand the isTTY part, can you clarify?

test/onPort.test.js
```diff
@@ -11,6 +11,7 @@ test('--on-port flag', { skip: !hasAsyncHooks() }, (t) => {
     /Running 1s test @ .*$/,
     /10 connections.*$/,
     /$/,
+    /$/,
```
mcollina, line 14: why?

test/envPort.test.js
```diff
@@ -10,6 +10,7 @@ const lines = [
   /Running 1s test @ .*$/,
   /10 connections.*$/,
   /$/,
+  /$/,
```
mcollina, line 13: why?

samples/track-run.js
```diff
@@ -13,6 +13,8 @@ function handle (req, res) {
 
 function startBench () {
   const instance = autocannon({
+    connections: 100,
+    duration: 2,
```
mcollina, line 17: why this change?

lib/manager.js
```diff
@@ -0,0 +1,71 @@
+const path = require('path')
+const os = require('os')
+var Spinner = require('cli-spinner').Spinner
+const { Worker } = require('./worker_threads')
+
+const spinner = new Spinner('running... %s')
+spinner.setSpinnerString('|/-\\')
+
+const numWorkers = Math.max(Math.floor(os.cpus().length * 0.75), 1)
```
mcollina, line 9: This should be an option/parameter, not decided based on the available CPUs. This should also be printed by the CLI.

lib/worker_threads.js
```diff
@@ -0,0 +1,9 @@
+const semver = require('semver')
```
mcollina, line 1: use strict.
Also avoid adding semver as a dependency. Try requiring `worker_threads` within a try-catch.

mcollina (changes requested):

test/onPort.test.js
salmanm: I had to add a [`\n` to the output](https://github.com/mcollina/autocannon/pull/290/files#diff-672f2640e048f0652357caf71e3a69e782159963c981fb5677cd06457e57d5b8R18) to make sure the result table isn't in the same line as spinner.

samples/track-run.js
salmanm: Leftover from my testing, haven't yet reviewed my PR.

lib/worker.js
salmanm: This was pre-existing in [autocannon.js](https://github.com/mcollina/autocannon/pull/290/files#diff-d19ffc6d9071d9610442b1d9173acd4c390dd5001bdc732bee7cd75e5f96aa36L260) when I used the same in worker.js. But it doesn't sound right to be in here. Removed.

lib/run.js
salmanm: Sorry not sure I understand what you mean. using `isMainThread` here ensures that workers do not try and initialize more workers. How would you do it?

lib/manager.js
salmanm: Just checked, it seems to be doing the same as it does on `master`, it's printing like this

image

autocannon.js
salmanm: Not sure I got it 100%, but updated to move runTracker in a separate file. LMK what you think

help.txt
```diff
@@ -15,6 +15,8 @@ Available options:
         The amount of requests to make before exiting the benchmark. If set, duration is ignored.
   -S/--socketPath
         A path to a Unix Domain Socket or a Windows Named Pipe. A URL is still required in order to send the correct Host header and path.
+  -W/--useWorkers
```
mcollina, line 18: I would call it `-t` and accept an integer (the number of threads)

lib/manager.js
mcollina: let's disable all spinners/progress bars for non-tty.

lib/printResult.js
```diff
@@ -0,0 +1,132 @@
...
+  ]
+}
+
+function asColor (colorise, row) {
+  return row.map((entry) => colorise(entry))
+}
+
+function asMs (stat) {
+  const result = Object.create(null)
+  Object.keys(stat).forEach((k) => {
```
mcollina, line 109: let's move these forEach to a for..of.

lib/run.js
mcollina: How is it possible for a worker to enter this function?

lib/runTracker.js
```diff
@@ -0,0 +1,31 @@
+const run = require('./run')
```
mcollina, line 1: use strict is missing

test/onPort.test.js
mcollina: We should really avoid this.

help.txt
salmanm: `t` is already being used for timeout and `T` for title.

mcollina: then let's use w but with an integer (number of threads)

lib/manager.js
salmanm: progressbar is already disabled for non-tty. Just did the same for spinner.
However, the above output (`[90m` `[39m`) isn't due to that.

lib/run.js
salmanm: worker calls this from [here](https://github.com/mcollina/autocannon/pull/290/files#diff-df9477f338c214ed04a5e566edd666461eee57595f4b8800d286075824c2ad2cR7).

mcollina (approved):
lgtm

simoneb:
LGTM. A couple of questions:

- as a user, how would I know when I should enable this option? Maybe some docs should be added about this? For instance: when you can't provide enough load and you see the CPU is the bottleneck, start using this option
- as a user, I would find it useful if this option had an "auto" setting, which uses an automatic heuristic about how many threads to use. It could still allow specifying an exact number though, as it's doing now. It would be a sort of "polymorphic" option whose value could be either "auto" or the number of threads to use

simoneb:
After trying it out, I have some additional comments:

- a UI one, see attached screenshot
  - when running without workers, I see a progress bar with no flickering
  - when running with workers, I see a spinner (no progress indicator) and the  "running" text flickers quite a bit

Untitled_ Oct 28, 2020 6_08 PM


- a stats one. The numbers don't seem to add up. I would expect the numbers to be the same between a run without or with workers, instead it seems that the req/sec and bytes/sec stats are averaged out instead of summed. See screenshot below

image

salmanm:
Good one @simoneb 

- About flickering, I think it might be terminal/os specific as I don't see any flickering in mac. 
- Good catch, I was relying on `hdr-histogram-js` to do the adding correctly, but not sure why it's `add` method averages the result. I'm looking into it.

simoneb:
> Good one @simoneb
> 
> * About flickering, I think it might be terminal/os specific as I don't see any flickering in mac.

Even though it's not a major issue as it doesn't affect functionality, I think it would still make sense to try to figure out why it's happening on other terminals. The point is, it provides a more professional look if it doesn't flicker and I would trust more a library which works and looks good. Lmk if you'd like some help on this, since I can reproduce it here.

simoneb (approved):
The aggregation issue I reported is solved.

I would still like to see:

- the spinner flickering go away
- an automatic option for worker threads and docs about it

Maybe these can be addressed in other PRs though

salmanm:
Quick benchmark result. 
Since single threaded autocannon is able to saturate CPU to 100% with one node process, I started a 3 worker node cluster and monitored CPU usage, each process wasn't going above 50%, which means we can now test the autocannon with workers.
I ran tests against it **3 times** for **10 seconds** with and without workers and got the following results

Without workers: 568k, 670k, 560k
With 4 workers: 801k, 817k, 793k
With 5 workers: 843k, 831k, 803k

Feels nice to finally be able to see the improvement. 🎉 

Btw with 6 workers the number started to drop again because of may be extra overhead.

mcollina:
> Btw with 6 workers the number started to drop again because of may be extra overhead.

How many CPUs do you have on your machine? You might be hitting some physical liimt.

salmanm:
I have 12 cores.

lib/manager.js
```diff
@@ -0,0 +1,136 @@
+'use strict'
+
+const path = require('path')
+const aggregateResult = require('./aggregateResult')
+const { getHistograms } = require('./histUtil')
+var Spinner = require('cli-spinner').Spinner
```
mcollina, line 6: Would you mind to use https://www.npmjs.com/package/ora instead? Thanks

lib/run.js
```diff
@@ -2,17 +2,46 @@
...
+function validateOpts (opts, cb) {
+  if (opts.workers) {
+    if (opts.setupClient) {
+      cb(new Error('Can not use setupClient option while using workers'))
+      return
+    }
+
+    if (opts.requests && opts.requests.some(r => r.setupRequest || r.onResponse)) {
+      cb(new Error('Can not use setupRequest/onResponse options while using workers'))
+    }
```
feugy, line 40: Can you give a bit more details about this limitation?

If I understand well, each worker is getting the same options (divided by the number of desired workers, when relevant).
If I ask 100 connections, with 5 workers, each will open 20 connections.

Why can't each connection invoke its own setupClient client, as it's doing when there is a single worker?

feugy (commented):
This looks absolutely great.

Thanks Salman! That was a difficult refactoring.
I've tried it on a live application, with duration, with amount, with number of connections... With and without numbers.

Matteo mentionned further work to enabled `setupClient` & `onResponse`, but I'm actually wondering why it is not possible already.


Oh, and can't we have a progress bar per worker?

troll

lib/run.js
salmanm: Data sent to workers uses structured clone algorithm, which allows only limited data types. See [here](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm).

> Function objects cannot be duplicated by the structured clone algorithm; attempting to throws a DATA_CLONE_ERR exception.

And since setupRequest/onResponse are functions, we might wanna come up with some clever idea to achieve the same.
Hence, we decided to postpone that work until this PR is merged.

salmanm:
@feugy progress bar per worker would be ideal. However, there may be too frequent worker/parent communication (when run using `--amount` arg) which I kinda wanted to avoid.

There are ways to simplify it too (like send update every second), but perhaps for later.

lib/run.js
```diff
@@ -2,17 +2,46 @@
...
+  if (opts.json) {
+    console.log(JSON.stringify(result))
+  }
+
+  // the code below this `if` just renders the results table...
+  // if the user doesn't want to render the table, we can just return early
+  if (!isMainThread || opts.renderResultsTable === false) return
+
+  printResult(result, opts)
+}
```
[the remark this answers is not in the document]
salmanm: I tried to keep the duplication to minimum, so it's this function call that is duplicate `printResult(result, opts)`. 
But would be happy to incorporate this change if you can suggest how would you simplify it.

mcollina (approved):
lgtm

Issue #296, opened 2020-11-06
Is it possible to render the table with Result ?


fraxken:
Hi,

I work in a project where we have automated the execution of several benchmarks and we keep all results in JSON format.

So I wondered afterwards if it was possible to render the Table by passing the JSON (JS) as argument? I would like to create a CLI to stdout the result of a benchmark at any time.

Thanks,
Best Regards,
Thomas

mcollina:
That would be fantatic but it's not currently supported! Would you like to send a Pull Request to address this issue? Remember to add unit tests.

fraxken:
Yes sure no problem, I will send a PR soon.

slaphead:
Thanks @fraxken for adding this. It works very well!

PR #297, opened 2020-11-09
Expose printResult internal

fraxken:
Hi,

The goal of this pull request is to expose the printResult internal function. I guess it need review, some of the things i was not sure:

- I created the printResult-process.js at the root of test (not sure if this is the right place).
- I used one of my result for fixtures/example-result (maybe the name and the content of the file need to be replaced).

I haven't tested all options yet.. I wanted to be sure the way i'm doing it is ok.

PR related to https://github.com/mcollina/autocannon/issues/296 

Best Regards,
Thomas


mcollina:
I would highly prefer to expose a function that returns a string and do not print directly to the console. This will require some changes to `printResult`.

fraxken:
Ok. I dedicated my Wednesday and Thursday on open-source this week. I will make an update to the PR at that time.

mcollina:
thanks!!!

fraxken:
Hello 👋

I started to make the necessary modifications. I added a new printResultAPI.js file in lib to avoid having to modify code directly in the autocannon root file.

For printResult i was not sure how to do it exactly
- I was thinking of using outputStream with a WritableStream but it would surely make printResultAPI Asynchronous (for no good reason i guess).
- One of my option was to return a string if outputStream was undefined or null (But I didn't necessarily want to introduce any inconsistency or break the existing code).

So finally i decided to go with a new parameter and a naïve solution and get your feedback on it.

mcollina:
I would prefer to have it always return a string.

fraxken:
Updated again. I was thinking you wanted to always return a string but only for API.

mcollina (commented):
Code is ok!

fraxken:
I removed the Draft status of the PR.

mcollina (approved):
lgtm

mcollina:
Good work!

PR #301, opened 2020-11-17
fix: handle large response body correctly

salmanm:
Supersedes https://github.com/mcollina/autocannon/pull/276/files
Fixes https://github.com/mcollina/autocannon/issues/275

Coincidently this fix also over-rides the code changes one done in https://github.com/mcollina/autocannon/pull/287 but the fix still works.

- Refactored response body parsing. It now collects body chunks until `kOnMessageComplete` fires and then it calls `onResponse` and emit `response`
- Added test to cover the same


mcollina (approved):
lgtm

PR #303, opened 2020-11-18
Adding support for passing in custom tlsOptions

chiragpat:
**Why**
Currently there is no way to pass custom client certificates through autocannon if the server being benchmarked requires client certificate validation

**How**
- Adding support in the js API to pass in tlsOptions through to the httpClient.
- Updated the test tls server to echo the client certificate email as a way to verify things are working correctly


mcollina (approved):
lgtm

Issue #279, opened 2020-08-31
setupRequest is being called before setupClient 


pola88:
Hi,

I'm having an issue, In the setupClient I assign some variables on the context, to use different values between the connections, that I have to use in the first request, but `setupRequest` is being called before `setupClient`, Should not be called `setupClient` first?

Thank in advance,
Pola.

mcollina:
I don't think the two are connected in any way. See https://github.com/mcollina/autocannon/blob/5112ca4375a49e6ae7f200a456951e73dd60b454/lib/httpClient.js#L34-L61 (`setupRequest` is called via `requestIterator`).

Can you please articulate what you are trying to do with a code example? Thanks

pola88:
Yes, sure, this is what I'm trying to do:

```
let connection = 0;
export const setupClientEventHandlers = (client) => {
  client.requestIterator.context = {
    user: seed[connection], //the seed has the email and password
    connection,
  };

  console.log(`Init connection ${connection}`);
  connection++;
};

const loginRequest = ({
  method: 'POST',
  path: '/sessions',
  name: 'login',
  setupRequest: (req, context) => ({
    ...req,
    body: JSON.stringify({
      email: `${context.user.email}`,
      password: `${context.user.password}`
    }),
  }),
  onResponse: function (status, body, context) {
    ...
  }
});

autocannon({
  url: url,
  duration: 60,
  timeout: 5,
  setupClient: setupClientEventHandlers,
  headers: {
    ...
  },
  requests: [
    loginRequest,
    ...
  ]
});
```

And this throw an error about `can not read property user on undefined` and also the log `Init connection` is not shown before the error. That why I think that setup request is not calling before the first request

Thanks in advance

mcollina:
I think you can just move https://github.com/mcollina/autocannon/blob/5112ca4375a49e6ae7f200a456951e73dd60b454/lib/httpClient.js#L34 after https://github.com/mcollina/autocannon/blob/5112ca4375a49e6ae7f200a456951e73dd60b454/lib/httpClient.js#L61.

Would you like to send a PR with a unit test?

dnlup:
This change makes a lot of sense, but it might break the usage of `setBody` and `setHeadersAndBody` inside `setupClient`:

* https://github.com/mcollina/autocannon/blob/master/test/httpClient.test.js#L202
* https://github.com/mcollina/autocannon/blob/master/test/httpClient.test.js#L491

salmanm:
[...]

PR #304, opened 2020-11-19
feat: Add support for initializing custom context object

salmanm:
Fixes https://github.com/mcollina/autocannon/issues/279

- Added `initialContext` option which is used to initialize context instead of empty object
- Covered with tests
- Updated readme.

### DIFF
--- a/lib/requestIterator.js
+++ b/lib/requestIterator.js
@@ -3,6 +3,7 @@
 const hyperid = require('hyperid')(true)
 const inherits = require('util').inherits
 const requestBuilder = require('./httpRequestBuilder')
+const clone = require('clone')
 
 function RequestIterator (opts) {
   if (!(this instanceof RequestIterator)) {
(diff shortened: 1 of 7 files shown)


lib/requestIterator.js
```diff
@@ -3,27 +3,33 @@
...
+  this.resetContext()
   this.reqDefaults = opts
   this.requestBuilder = requestBuilder(opts)
   this.setRequests(opts.requests)
 }
 
 inherits(RequestIterator, Object)
 
+RequestIterator.prototype.resetContext = function () {
+  this.context = clone(this.initialContext)
```
mcollina, line 24: please use the `rdfc` module instead.

mcollina (changes requested):

lib/requestIterator.js
salmanm: Aye, done. Also replaced another pre-existing use of `clone`.

mcollina (approved):
lgtm

lib/requestIterator.js
salmanm: Turns out, `rfdc` doesn't support copying `Buffer`s. I'll take that separately. Reverted the last commit.

mcollina:
can you fix the conflicts?

PR #308, opened 2020-11-30
fix: add workers in the output

salmanm:
- Added workers option in the output
- Added a test for workers


mcollina (approved):
lgtm

salmanm:
Damn these windows tests aren't predictable. Increased timeout to 8 mins for now. Are we ok with it or should investigate why this happens?

PR #310, opened 2020-12-07
fix: fixes (mem leaks, progress tracking)

salmanm:
- Fixed event emitter memory leaks
- Fixed progress tracker for workers mode

### DIFF
--- a/.taprc
+++ b/.taprc
@@ -2,5 +2,5 @@ esm: false
 ts: false
 jsx: false
 flow: false
-timeout: 500
+timeout: 900
 coverage: true
--- a/lib/init.js
+++ b/lib/init.js
@@ -38,7 +38,7 @@ function _init (opts, tracker, cb) {
       return cbPassedIn ? cb(err) : setImmediate(() => tracker.emit('error', err))
     }
 
-    setImmediate(() => tracker.emit('done', result))
+    tracker.emit('done', result)
     cb(null, result)
 
     if (!err && opts.json) {
--- a/lib/runTracker.js
+++ b/lib/runTracker.js
@@ -10,10 +10,6 @@ function runTracker (opts, tracker, cb) {
 
   run(opts, tracker, cb)
 
-  process.once('SIGINT', () => {
-    tracker.stop()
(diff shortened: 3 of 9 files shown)


lib/init.js
```diff
@@ -38,7 +38,7 @@ function _init (opts, tracker, cb) {
       return cbPassedIn ? cb(err) : setImmediate(() => tracker.emit('error', err))
     }
 
-    setImmediate(() => tracker.emit('done', result))
```
salmanm, line 41: `setImmediate` is not needed here because unlike `error` case, we'll reach at this point only after enough delay so that the consumer would get time to bind the events.

and the for removing is so `done` fires instantly and progress tracker gets a chance to clear the cli progress/spinner **before** anyone writes anything in the output.

lib/runTracker.js
```diff
@@ -10,10 +10,6 @@ function runTracker (opts, tracker, cb) {
 
   run(opts, tracker, cb)
 
-  process.once('SIGINT', () => {
```
salmanm, line 13: This was the cause for mem leak warning in the console.
It wasn't a use-case issue though. It was coming from wasm.test.js where we're running LOTS of parallel tests and each one was binding this SIGINT event.

.taprc
```diff
@@ -2,5 +2,5 @@ esm: false
 ts: false
 jsx: false
 flow: false
-timeout: 500
+timeout: 900
```
salmanm, line 5: May be this is not ideal, but in [some tests](https://github.com/mcollina/autocannon/blob/master/test/httpClient.test.js#L580) we're waiting for a couple of seconds which may be messing up with tap timeouts on windows.

salmanm: Fails after 3 mins even though the timeout is 15 mins now. I'll investigate more.

salmanm: Splitting some tests into multiple blocks in test/runRate.test.js seems to have fixed the issue.
_Or may be it's appearing as so._ 😅

mcollina (approved):
lgtm

PR #313, opened 2020-12-09
fix: attempt to fix intermittent test failures

salmanm:
Builds are still not very stable on windows CI.
Although we tried running on windows machine locally but could never see a test fail.

### In `httpClient.test.js`

It seems like these two tests in `httpClient.test.js` are failing intermittently. 

- client should emit a timeout when no response is received
- client should emit 2 timeouts when no responses are received

This PR increases wait time from 2500 to 2800, just so slow windows boxes don't choke up. 

### in `runRate.test.js`

All 3 scenarios of the following test are failing intermittently. 

- run should only send the expected number of requests per second

I presume the `duration` may be coming as slightly more than 4.5 seconds due to slow windows box and becomes `5` due to `Math.round`. But we're asserting it with `4`, so I've changed it to `Math.floor` instead.

### DIFF
--- a/test/httpClient.test.js
+++ b/test/httpClient.test.js
@@ -563,7 +563,7 @@ test('client should emit a timeout when no response is received', (t) => {
     t.ok(1, 'timeout should have happened')
   })
 
-  setTimeout(() => client.destroy(), 1500)
+  setTimeout(() => client.destroy(), 1800)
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm, might be the right time!

Issue #315, opened 2020-12-14
Avg latency is wrong when using pipelining


simoneb:
This was apparently addressed in a very old issue #15 but I am still seeing something wrong with latency when pipelining is used, especially in AVG. This is unless I'm interpreting the number wrong. I assume latency is the time a request takes to complete, as described in the docs.

Here's my web app:

```
const http = require("http");

http
  .createServer(function (req, res) {
    setTimeout(function () {
      res.end("hello world");
    }, 500);
  })
  .listen(3000);
```

And here's autocannon's output without and then with pipelining:

```
# without pipelining

autocannon http://localhost:3000
Running 10s test @ http://localhost:3000
10 connections

┌─────────┬────────┬────────┬────────┬────────┬───────────┬─────────┬────────┐
│ Stat    │ 2.5%   │ 50%    │ 97.5%  │ 99%    │ Avg       │ Stdev   │ Max    │
├─────────┼────────┼────────┼────────┼────────┼───────────┼─────────┼────────┤
│ Latency │ 499 ms │ 513 ms │ 522 ms │ 523 ms │ 510.27 ms │ 6.47 ms │ 523 ms │
└─────────┴────────┴────────┴────────┴────────┴───────────┴─────────┴────────┘
```

```
# with pipelining

autocannon http://localhost:3000 -p 2
Running 10s test @ http://localhost:3000
10 connections with 2 pipelining factor

┌─────────┬──────┬──────┬────────┬────────┬───────────┬───────────┬────────┐
│ Stat    │ 2.5% │ 50%  │ 97.5%  │ 99%    │ Avg       │ Stdev     │ Max    │
├─────────┼──────┼──────┼────────┼────────┼───────────┼───────────┼────────┤
│ Latency │ 0 ms │ 0 ms │ 527 ms │ 528 ms │ 256.62 ms │ 256.65 ms │ 530 ms │
└─────────┴──────┴──────┴────────┴────────┴───────────┴───────────┴────────┘
```

As you can see, the AVG latency when using pipelining seems to be divided by the pipelining factor, which doesn't sound right to me, in the sense that the server is still responding in ~500ms, so the average latency should always be around that value. Possibly as a consequence of that, Stdev also becomes wrong.

mcollina:
Good spot, PR? I _never_ look at the avg latency, just at 99% percentile.

PR #317, opened 2020-12-15
fix (#315): correctly increment this.cer

salmanm:
Fixes https://github.com/mcollina/autocannon/issues/315

### DIFF
--- a/lib/httpClient.js
+++ b/lib/httpClient.js
@@ -112,6 +112,9 @@ function Client (opts) {
 
     this.emit('response', resp.headers.statusCode, resp.bytes, resp.duration, this.rate)
 
+    // Separating rip ensures that after getting response for first request, we fire
+    // first one again because second is already running right now (with pipelining=2).
+    // See detailed explaination at https://github.com/mcollina/autocannon/pull/317#discussion_r543678831
     const rpi = (this.cer + pipelining) % pipelining
     resp.body = ''
     resp.bytes = 0
--- a/lib/httpClient.js
+++ b/lib/httpClient.js
@@ -112,10 +112,11 @@ function Client (opts) {
 
     this.emit('response', resp.headers.statusCode, resp.bytes, resp.duration, this.rate)
 
+    const rpi = (this.cer + pipelining) % pipelining
     resp.body = ''
     resp.bytes = 0
     this.cer = (this.cer + 1) % pipelining
-    this._doRequest(this.cer)
+    this._doRequest(rpi)
   }
 
   this._connect()
--- a/lib/httpClient.js
+++ b/lib/httpClient.js
@@ -114,7 +114,7 @@ function Client (opts) {
 
     resp.body = ''
     resp.bytes = 0
-    this.cer = this.cer === pipelining - 1 ? 0 : this.cer++
+    this.cer = (this.cer + 1) % pipelining
     this._doRequest(this.cer)
   }
 
--- a/test/httpClient.test.js
+++ b/test/httpClient.test.js
@@ -63,6 +63,25 @@ test('client calls a https server twice', (t) => {
   })
 })
 
+test('client calculates correct duration when using pipelining', (t) => {
+  t.plan(4)
+
+  const lazyServer = helper.startServer({ delayResponse: 500 })
+  const opts = lazyServer.address()
+  opts.pipelining = 2
+  const client = new Client(opts)
+  let count = 0
+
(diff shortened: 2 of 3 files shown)


mcollina (approved):
lgtm

mcollina:
Actually we need tests

lib/httpClient.js
```diff
@@ -112,10 +112,11 @@ function Client (opts) {
 
     this.emit('response', resp.headers.statusCode, resp.bytes, resp.duration, this.rate)
 
+    const rpi = (this.cer + pipelining) % pipelining
```
salmanm, line 115: rpi = request pipelining index
cer = current expected response

Separated `rpi` and `cer` because all pipelining requests are fired at once without waiting for response and the following happens (with pipelining = 2)

- Both requests fired
- `cer` is **0**
- Response for first came
- After processing it, we call `_doRequest` with incremented `cer`
- Incremented `cer` is now **1** 
- `_doRequest` tries to fire the same request again, hence re-setting the startTime to now which messes up the `duration` calculation when response comes.

Separating `rip` ensures that after getting response for **first** request, we fire **first** one again because second is already running right now.

Does it make sense? 😅

mcollina: Totally, put it in a comment ;).

GlenTiki: All of this code was written at a time when I understood function character length had a direct impact on it being optimised. Newer versions of v8 are not limited by this afaik. If I was refactoring this code now I would recommend rewriting the variables as their full names, and adding many, many more comments.

`rpi` = `requestPipeliningIndex`
`cer` = `currentExpectedResponse`

After looking back at this code for the first time in years, I was struggling to understand the variable meanings myself. I would even being to rename many functions (even anonymous ones), to give the code more contextual meaning. I'd aim to make it readable/understandable first, then measure the impact. My biggest mistake writing the code in autocannon was not doing that during initial implementations, so I'm advising to learn from that 😅 

I am happy to refactor this in a follow up PR, if you'd like.

test/httpClient.test.js
```diff
@@ -63,6 +63,25 @@ test('client calls a https server twice', (t) => {
...
+
+  const lazyServer = helper.startServer({ delayResponse: 500 })
+  const opts = lazyServer.address()
+  opts.pipelining = 2
+  const client = new Client(opts)
+  let count = 0
+
+  client.on('response', (statusCode, length, duration) => {
+    t.equal(statusCode, 200, 'status code matches')
+    t.ok(duration > 500 && duration < 600)
```
simoneb, line 77: this is probably going to make it a brittle test at some point. I understand that it's tricky to test though, but relying on timing in tests is something I would try to avoid. I am not sure how to do it though, apart from changing the way the test is written altogether

lib/httpClient.js
simoneb: @salmanm your explanation above makes sense to me, I just want to make sure there isn't an implicit assumption in this statement:

> Separating rip ensures that after getting response for first request, we fire first one again because second is already running right now.

This is not assuming that requests will take similar times to execute, right? In other words, this supports the case whereby, with pipelining 2:

- fire first request
- fire second request
- **second  response comes back** before the first
- first response comes back

test/httpClient.test.js
salmanm: We're doing such time based assertions in a lot of places in these tests, so I don't think this is any less appropriate than that. But if you have any clever idea on how to improve, I'd be happy to hear it.

salmanm:
> This is not assuming that requests will take similar times to execute

Not at all, it's based on the fact that the pipelining logic has already fired a number of requests in the beginning and we just wanna avoid making redundant calls.

Yes, it supports the case you described. Besides, the existing logic does not seem to care if second response came first. See `cer` variable. It already assumes the current expected response. Which isn't 100% precise, but is fine IMO.

test/httpClient.test.js
simoneb: Not really, I think the only improvement to reduce possible brittleness is to only assert that it will take longer than 500, because setTimeout is guaranteed to trigger after and not before the specified time

salmanm: our calculations were giving ~half results, so not checking if it takes less than 600ms might not catch if the calculations in future break and start giving 2x time. We can increase it to 700 if builds start to fail.

simoneb: fair enough!

simoneb (approved):

PR #318, opened 2020-12-18
fix: test checking duration of request with pipeline

salmanm:
image

Tests are failing on MacOS 10. May be 600 ms isn't enough time, so increased it to 800


mcollina (approved):
lgtm

Issue #324, opened 2021-01-20
TypeError: Worker is not a constructor


Xiryl:
Hi,
I'm using the library with the worker options, using the `onResponse` as file module as said on the documentation:

```node
const runWrkPOST = (connections, pipelining, duration, title) => {
  const instance = autocannon(
    {
      url: 'http://localhost:3000',
      duration: 2,
      title,
      workers: 2,
      headers: {
        'Content-type': 'application/json; charset=utf-8',
      },
      requests: [
        {
          method: 'POST',
          path: '/api/v1/k/push/example',
          onResponse: path.join(__dirname, 'helpers', 'on-response'),
        },
      ],
    },
    (err, result) => console.log(result),
  );
  autocannon.track(instance, {
    renderProgressBar: true,
    renderResultsTable: true,
    renderLatencyTable: true,
  });
};
```

where in `./helpers/on-response.js`:
```node
const fastJson = require('fast-json-stringify');

module.exports = (client, statusCode, resBytes, responseTime) => {
  const stringify = fastJson({
    title: 'Message Schema',
    type: 'object',
    properties: {
      message: {
        type: 'string',
      },
    },
  });
  client.setHeaders({ 'Content-type': 'application/json; charset=utf-8' });
  client.setBody(
    stringify({
      message: `RESTAPI,datatype=Double,unit=°C ATemp=19.19 ${new Date().getTime()}000000`,
    }),
  );
};
```

But I'm getting the following error:
```sh
/Users/fabio/projects/internal/test/test/node_modules/autocannon/lib/manager.js:69
      const worker = new Worker(path.resolve(__dirname, './worker.js'), { workerData: { opts: workerOpts } })
                     ^

TypeError: Worker is not a constructor
    at init (/Users/fabio/projects/internal/test/test/node_modules/autocannon/lib/manager.js:69:22)
    at initWorkers (/Users/fabio/projects/internal/test/test/node_modules/autocannon/lib/manager.js:107:3)
    at _init (/Users/fabio/projects/internal/test/test/node_modules/autocannon/lib/init.js:57:5)
    at init (/Users/fabio/projects/internal/test/test/node_modules/autocannon/lib/init.js:13:10)
    at Object.runWrkPOST (/Users/fabio/projects/internal/test/test/src/bapi.js:61:20)
    at Object.<anonymous> (/Users/fabio/projects/internal/test/test/index.js:22:6)
    at Module._compile (internal/modules/cjs/loader.js:778:30)
[...]

PR #325, opened 2021-01-20
fix(324): Show error when no worker support is found

salmanm:
Fixes https://github.com/mcollina/autocannon/issues/324

### DIFF
--- a/test/run.test.js
+++ b/test/run.test.js
@@ -366,7 +366,7 @@ test('run should accept a unix socket/windows pipe', (t) => {
 })
 
 for (let i = 1; i <= 5; i++) {
-  test(`run should count all ${i}xx status codes`, (t) => {
+  test(`run should count all ${i}xx status codes`, { skip: true }, (t) => {
     const server = helper.startServer({ statusCode: i * 100 + 2 })
 
     initJob({
--- a/test/run.test.js
+++ b/test/run.test.js
@@ -3,10 +3,12 @@
 const os = require('os')
 const path = require('path')
 const test = require('tap').test
+const semver = require('semver')
 const initJob = require('../lib/init')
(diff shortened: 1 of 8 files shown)


test/run.test.js
```diff
@@ -366,7 +366,7 @@ test('run should accept a unix socket/windows pipe', (t) => {
 })
 
 for (let i = 1; i <= 5; i++) {
-  test(`run should count all ${i}xx status codes`, (t) => {
+  test(`run should count all ${i}xx status codes`, { skip: true }, (t) => {
```
salmanm, line 369: Has to disable this temporarily as it fails consistently only on node 15.6.0.
We may need to investigate this separately.

mcollina: Can you please make it skip only on node v15?

mcollina (changes requested):

test/run.test.js
salmanm: Done

test/run.test.js
```diff
@@ -366,7 +368,7 @@ test('run should accept a unix socket/windows pipe', (t) => {
 })
 
 for (let i = 1; i <= 5; i++) {
-  test(`run should count all ${i}xx status codes`, (t) => {
+  test(`run should count all ${i}xx status codes`, { skip: isNode15 }, (t) => {
```
mcollina, line 371:
```suggestion
  // TODO we should not skip those tests
  test(`run should count all ${i}xx status codes`, { skip: isNode15 }, (t) => {
```

mcollina (commented):
tiny nit and we can land

test/run.test.js
salmanm: Sure, I'm actually gonna look into the issue right after 🙂

mcollina (approved):

PR #328, opened 2021-01-21
fix: broken test with node 15

salmanm:
```js
  const server = http.createServer((req, res) => {
    res.statusCode = 102
    res.end('Hello')
  })
  server.listen(9000)
```

This simple http server responds differently in Node v14 vs v15. 

**Response in Node v14**

```
HTTP/1.1 102 Processing
Date: Thu, 21 Jan 2021 15:22:09 GMT
Connection: close
```

**Response in Node v15**

```
HTTP/1.1 102 Processing
Date: Thu, 21 Jan 2021 15:22:09 GMT
Connection: close
Content-Length: 5 // <---
```

I haven't done a thorough research with node and http specs as to why this behaviour is different but as I understand, the `102` response doesn't return a body. But if you try to send it, Node 14 smartly ignores it and does not send a `Content-Length` header. However Node 15 sends this header, which is why `http-parser-js` waits for the body and never fires `kOnMessageComplete`.

### DIFF
--- a/test/helper.js
+++ b/test/helper.js
@@ -36,7 +36,10 @@ function startServer (opts) {
     }
 
     res.statusCode = statusCode
-    const reply = () => res.end(typeof body === 'function' ? body(req) : body)
+    const reply = () => {
+      const bodyToWrite = typeof body === 'function' ? body(req) : body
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

PR #332, opened 2021-02-16
fix: Reuse headers when set - fixes #280

### LINKED ISSUE #280: client.setHeaders only sets headers for first request
I'm not sure if this is intended behavior or a bug, but in the `setupClient`, while doing a `client.setHeaders`, it only affects the first request in the `requests` array.

Example:

```js
'use strict'

const autocannon = require('autocannon')

autocannon({
    url: `https://example.com`,
    connections: 1,
    duration: 1,
    amount: 1,
    setupClient: (client) => {
        const userInfo = getNextUserInfo()
        // I expect this to set the headers for all requests in the list below
        client.setHeaders({
            'Authorization': userInfo.authorizationToken,
            'Content-Type': 'application/json',
        })
    },
    requests: [
        {
            path: `/someEndpoint`,
            method: 'POST',
            setupRequest: (req, context) => {
                // I do this hack to get the headers forwarded to the next request;
                // context.authHeaders = req.headers // left out to show bug
                return {
                    ...req,
                    body: JSON.stringify({
                        command: 'startTournamentLevel',
                    }),
                }
            },
            onResponse: (status, body, context) => {
                if (status === 200) {
                    context.data = JSON.parse(body)
                }
            },
        },
        {
            path: `/someOtherEndpoint`,
            method: 'POST',
            setupRequest: (req, context) => {
                console.log('req.headers is not set', JSON.stringify(req.headers))
                return {
                    ...req,
                    // headers: context.authHeaders, // I do this hack to get the headers forwarded. Left out to expose bug
                    body: JSON.stringify({
                        someData: context.data.some,
                    }),
                }
            },
        }],
}, console.log)
```

I think I have found the reason in the source code;

The method on the `httpClient` passes the to the requestIterator

https://github.com/mcollina/autocannon/blob/53643534b935e81970244ea1814cd74572b9e17d/lib/httpClient.js#L211

It seems this method only sets the headers for the `currentRequest`
https://github.com/mcollina/autocannon/blob/53643534b935e81970244ea1814cd74572b9e17d/lib/requestIterator.js#L54

I'm not sure what is the correct behavior, but I would expect that perhaps the `httpClient` saved the headers, and added it to all requests as the new default.

Again, not sure if it is a bug, but it feels like it. I was able to work around it by adding the headers to the context object on the first request, in case anyone else has this issue.

SimonJang:
Thanks for creating and maintaining `autocannon`. I really enjoy using this software.

I was running tests on an AWS API Gateway where I needed to configure a fixed set of headers for a couple of requests when I encountered the same issue as in #280 .

I went ahead and tried to fix it. All feedback is welcome and I'll try to update as quick as possible.

### DIFF
--- a/lib/requestIterator.js
+++ b/lib/requestIterator.js
@@ -11,6 +11,7 @@ function RequestIterator (opts) {
   }
 
   this.resetted = false
+  this.headers = {}
   this.initialContext = opts.initialContext || {}
   this.resetContext()
   this.reqDefaults = opts
@@ -61,7 +62,8 @@ RequestIterator.prototype.setRequests = function (newRequests) {
 }
 
 RequestIterator.prototype.setHeaders = function (newHeaders) {
-  this.currentRequest.headers = newHeaders || {}
+  this.headers = newHeaders || {}
+  this.currentRequest.headers = this.headers
   this.rebuildRequest()
 }
 
@@ -71,7 +73,8 @@ RequestIterator.prototype.setBody = function (newBody) {
 }
 
 RequestIterator.prototype.setHeadersAndBody = function (newHeaders, newBody) {
-  this.currentRequest.headers = newHeaders || {}
+  this.headers = newHeaders || {}
+  this.currentRequest.headers = this.headers
   this.currentRequest.body = newBody || Buffer.alloc(0)
   this.rebuildRequest()
 }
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

PR #340, opened 2021-03-01
Remove unnecessary timeout check in httpClient delay test

nherment:
In addition it's probably better to use a t.setTimeout(..) than assertion for a timeout because it will come with a clear and informative error message.


mcollina (approved):
lgtm

mcollina:
Keep them coming!

PR #341, opened 2021-03-02
Update some more tests relying on timeouts

nherment:
@mcollina this should help further.

However, I don't like the fix here because as a user of `httpClient` I would expect for `client.destroy()` to work regardless of when it is called. In the current implementation, `client.destroy()` will not work as expected if invoked synchronously in a timeout handler. It does not clear the retry mechanism of the `httpClient`.

I think the "right" fix consists in using a different version of `retimer` that allows to reset the timer after it has been triggered. This way there is no need to change the instance of retimer within the `httpClient`.

Cf. https://github.com/mcollina/autocannon/blob/f8b598e103644e5048f9a936d6ac3b746986285e/lib/httpClient.js#L73

### DIFF
--- a/test/httpClient.test.js
+++ b/test/httpClient.test.js
@@ -581,9 +581,7 @@ test('client should emit a timeout when no response is received', (t) => {
 
   client.on('timeout', () => {
     t.ok(1, 'timeout should have happened')
-
-    // client.destroy must be done async to ensure the correct internal timer is destroyed instead of the one that triggered this timeout
-    setTimeout(() => client.destroy())
+    client.destroy()
   })
 })
 
@@ -597,8 +595,7 @@ test('client should emit 2 timeouts when no responses are received', (t) => {
   client.on('timeout', () => {
     t.ok(1, 'timeout should have happened')
     if (count++ > 0) {
-      // client.destroy must be done async to ensure the correct internal timer is destroyed instead of the one that triggered this timeout
-      setTimeout(() => client.destroy())
+      client.destroy()
     }
   })
 })
--- a/test/httpClient.test.js
+++ b/test/httpClient.test.js
@@ -581,9 +581,10 @@ test('client should emit a timeout when no response is received', (t) => {
 
   client.on('timeout', () => {
     t.ok(1, 'timeout should have happened')
+    
+    // client.destroy must be done async to ensure the correct internal timer is destroyed instead of the one that triggered this timeout
+    setTimeout(() => client.destroy())
   })
-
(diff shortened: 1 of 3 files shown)


mcollina:
wooow! Good spot. Could you fix retimer? Another way might be to just drop its usage. The cost of allocating a timer almost disappeared in latest Node.js

nherment:
It makes sense to update retimer as it does add some functionality over the standard setTimeout().

nherment:
@mcollina the Pr has been updated with the latest retimer and the code updated.

mcollina (approved):
lgtm

test/httpClient.test.js
```diff
@@ -66,15 +66,16 @@ test('client calls a https server twice', (t) => {
...
+  const lazyServer = helper.startServer({ delayResponse })
   const opts = lazyServer.address()
   opts.pipelining = 2
   const client = new Client(opts)
   let count = 0
 
   client.on('response', (statusCode, length, duration) => {
     t.equal(statusCode, 200, 'status code matches')
-    t.ok(duration > 500 && duration < 800)
+    t.ok(duration > delayResponse, `Expected response delay > ${delayResponse}ms but got ${duration}ms`)
```
salmanm, line 78: FYI the original reason for checking if duration < 800 was [this](https://github.com/mcollina/autocannon/pull/317#discussion_r544185281).

> our calculations were giving ~half results, so not checking if it takes less than 600ms might not catch if the calculations in future break and start giving 2x time. We can increase it to 700 if builds start to fail.

I know it wasn't the right fix and there should be a more predictable way to assert this but removing the upper limit check brings the risk I explained in the comment.

mcollina: @nherment could you handle this?

nherment: Yes, I'll add a unit test for these calculations.

nherment: @salmanm could you please point me to the 'calculations' you are referring to?

The code around the delayed response is pretty straightforward and I don't see anything that I would classify as calculations.

salmanm: Calcs were wrong for avg latency when using pipeline (See [this](https://github.com/mcollina/autocannon/issues/315)). Which was fixed by correctly calculating current expected response (`cer`) [here](https://github.com/mcollina/autocannon/pull/317/files#diff-9c6cf685b37a3aab04dbaf56e9a8d4bfb60041335227154d65f0d69d9c2f7b23R118).

PR #342, opened 2021-03-03
fix: do not show warnings when workers >= 10

### LINKED ISSUE #337: Setting workers to >= 10 yields eventemitter warnings
```
root@DESKTOP-5KK9VIR:~/node/node# autocannon -w 10 localhost:3000
Running 10s test @ http://localhost:3000
10 connections
10 workers

⠋ running...(node:9067) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11
 unpipe listeners added to [WriteStream]. Use emitter.setMaxListeners() to increase limit
(Use `node --trace-warnings ...` to show where the warning was created)
(node:9067) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 error liste
ners added to [WriteStream]. Use emitter.setMaxListeners() to increase limit
(node:9067) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 close liste
ners added to [WriteStream]. Use emitter.setMaxListeners() to increase limit
(node:9067) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 finish list
eners added to [WriteStream]. Use emitter.setMaxListeners() to increase limit
⠦ running...
```

simoneb:
Replaces ora with char-spinner to avoid the warning altogether.

### old explanation

I didn't actually figure out the root cause of why this is happening, and I do realize that making this initialization eager is doing something that's probably unnecessary unless workers are being used. Nevertheless, that's the most reliable way I found to fix this.

The problem seems to come from the bl package used by ora, which does some fancy stream manipulation, eventually leading to the warnings being generated.

Making the initialization of ora eager in the module root makes the issue disappear. 

Closes #337

### DIFF
--- a/lib/progressTracker.js
+++ b/lib/progressTracker.js
@@ -8,7 +8,7 @@
 const ProgressBar = require('progress')
 const Chalk = require('chalk')
 const testColorSupport = require('color-support')
-const ora = require('ora')
+const charSpinner = require('char-spinner')
 const printResult = require('./printResult')
 const { isMainThread } = require('./worker_threads')
 const defaults = {
@@ -79,13 +79,12 @@ function track (instance, opts) {
   })
 
   function showSpinner () {
-    spinner = ora('running...')
-    spinner.start()
+    spinner = charSpinner()
   }
 
   function hideSpinner () {
     if (spinner) {
-      spinner.stop()
+      clearInterval(spinner)
       spinner = null
     }
   }
--- a/package.json
+++ b/package.json
@@ -42,12 +42,14 @@
     "busboy": "^0.3.1",
     "pre-commit": "^1.1.2",
     "proxyquire": "^2.1.3",
+    "sinon": "^9.2.4",
     "split2": "^3.2.2",
     "standard": "^16.0.3",
     "tap": "^14.10.8"
   },
   "dependencies": {
     "chalk": "^4.1.0",
+    "char-spinner": "^1.0.1",
     "cli-table3": "^0.6.0",
     "clone": "^2.1.2",
     "color-support": "^1.1.1",
@@ -61,7 +63,6 @@
     "manage-path": "^2.0.0",
     "minimist": "^1.2.0",
     "on-net-listen": "^1.1.1",
-    "ora": "^5.1.0",
     "pretty-bytes": "^5.4.1",
     "progress": "^2.0.3",
     "reinterval": "^1.1.0",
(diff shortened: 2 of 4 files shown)


lib/progressTracker.js
```diff
@@ -18,6 +18,7 @@ const defaults = {
   renderResultsTable: true,
   renderLatencyTable: false
 }
+const spinner = ora('running...')
```
mcollina, line 21: Can you please instantiate this on first use instead?

simoneb: that will reintroduce the problem again :)

mcollina: Keeping it global and instantiating it only on first use should be identical to this.

simoneb: it is not, see the change I did which delays initialization and see the test I added failing. _when_ it is initialized is what makes the difference and that's also part of the reason why I couldn't quite figure out the root cause

lib/progressTracker.js
```diff
@@ -34,7 +35,6 @@ function track (instance, opts) {
   let durationProgressBar
   let amountProgressBar
   let addedListeners = false
-  let spinner
```
simoneb, line 37: even if I move the instantiation here, it will still cause the test failure. moving it outside this method is what fixes it

lib/progressTracker.js
simoneb: ok no wait a second, there's something gnarly at play here. the test is failing with this version of the code on my machine, but it is not failing on CI. looking into that

simoneb: ok the reason was unrelated. the test file for the progress tracking where I added this test was wrongly named and its tests were never run (I was running that file directly when developing locally, hence I was I saw it failing)

salmanm:
I think the problem starts when Ora tries to mute `process.stdout` stream. And `process.stdout._events.xxx` already contains `n` (number of active workers) listeners. So I appreciate that it's not straightforward.

But why don't we simply get rid of Ora!? I know it's widely used but if it's causing problems, I think it's worth looking for alternatives. `char-spinner` for instance, it's very old and not actively maintained, but I think it works fine and a very simply code base. I'd say we start using it, and if we need to tweak, we can just fork it.

simoneb:
@salmanm fair point. I changed the implementation. rationale was, I'd spent quite some time looking into the issue and the solution seemed good enough. char-spinner is much simpler it turns out so it's a more robust solution indeed. I'd assumed there was some thought put into choosing the spinner library so I didn't consider reverting that decision

mcollina (approved):
lgtm

PR #343, opened 2021-03-03
Ensure there are tests for calculating the duration of pipelined requests

nherment:
This PR mostly brings 2 changes:

1. It extracts the previous [cer/rpi](https://github.com/mcollina/autocannon/pull/317/files#r543678831) logic into a specialized queue`PipelinedRequestsQueue`.
2. It adds asserts and tests to ensure the reported response time is bound between the induced delay and the observed response time, from the test's perspective


lib/pipelinedRequestsQueue.js
```diff
@@ -0,0 +1,78 @@
+/**
```
mcollina, line 1: Please add a `'use strict'` at the top.

mcollina (commented):
Good work!

mcollina (approved):
lgtm

PR #345, opened 2021-03-04
fix: suppress test output

simoneb:
- restore original progressTracker test stub which was not intended to be run automatically
- move test which checks warnings into own module and run it with suppressed output


mcollina (approved):
lgtm

Issue #346, opened 2021-03-04
Add display of all response status codes, not just 2xx vs non-2xxx


maxfortun:
When load testing authorization it is very useful to see 30xs, 401s, and 403s.
When load testing proxies it is useful to see 502s and 504s.

Fdawgs:
closed by #347

PR #347, opened 2021-03-04
Adding individual status code statistics table

maxfortun:
#346 
Add `--renderStatusCodes` flag to render individual status code statistics table 

```
┌──────┬───────┐
│ Code │ Count │
├──────┼───────┤
│ 401  │ 124   │
└──────┴───────┘
```


mcollina (changes requested):
would you mind to add a unit test?

lib/run.js
```diff
@@ -216,6 +219,13 @@ function run (opts, tracker, cb) {
       tracker.emit('response', this, statusCode, resBytes, responseTime)
       const codeIndex = Math.floor(parseInt(statusCode) / 100) - 1
       statusCodes[codeIndex] += 1
+
+      if (!statusCodeStats[statusCode]) {
+        statusCodeStats[statusCode] = { count: 1 }
+      } else {
+        statusCodeStats[statusCode].count++
+      }
```
simoneb, line 227: is there any reason to not use the object values as the count, instead of having an object with a single property `count`? It would be more consistent with how the other stats are aggregated

simoneb (commented):
There's a room for an overall improvement here. Status codes are already being aggregated, but they're aggregated by "index": 2xx, 3xx, ..., which is less granular than what you're trying to achieve with this change. Instead of creating an additional implementation which does the same but in a more granular way, why don't you make the existing implementation more granular and use the finer grained granularity when `renderStatusCode` is true, and fallback to the existing behavior when it's false?

lib/run.js
maxfortun: @simoneb , you are absolutely right, there are many ways this could be improved. I was not certain of how far I can mod the code and still remain within the original author's intent. My goal was `additive` change only without impacting any of the existing functionality. If you believe that you can take this further, please do.

maxfortun: As far as `object with a  single property`, I envision that there may be other attributes aggregated about each individual statusCode, other than a count, and thus wanted to provide a harness for that. An example could be request timing average per code.

simoneb: @maxfortun on the overall approach let's hear what @mcollina thinks. On the count property, if it's not necessary, I'd say avoid it

maxfortun:
@mcollina unit test added

mcollina (approved):
lgtm

mcollina (approved):
lgtm

test/printResult-process.js
```diff
@@ -2,6 +2,14 @@
...
 
-const resultStr = autocannon.printResult(exampleResult)
+let opts = null
+
+if (process.argv.length > 2) {
+  const args = crossArgv(process.argv.slice(2))
+  opts = autocannon.parseArguments(args)
+}
+
+const resultStr = autocannon.printResult(exampleResult, opts)
```
mcollina, line 14: I'm not sure this change to the tests is actually being used anywhere. Can you please confirm?

mcollina (changes requested):

test/printResult-process.js
maxfortun: Looks like I forgot to add a file. Fixed. Thanks.

mcollina (approved):
lgtm

Issue #202, opened 2019-05-16
Add support for a warmup interval before sampling


GlenTiki:
This would mean that requests made in the first x amount of time during the warmup interval are not sampled. This should allow the server being tested to stabilise with the traffic from autocannon before sampling its load capabilities.

afbarbaro:
I would upvote this request.  We're using [nextjs](https://nextjs.org/) and it does several things right as it boots (like compiling pages) that affect performance.  Having the ability to delay the sampling start will be key to get good measurements.
Thank you!

simoneb:
How would you expect this to work? What pattern of requests should the warmup stage execute?

mcollina:
Same as the other one.

simoneb:
Doesn't it get tricky when the run parameters are (as they are by default) duration-based? Do you run the warmup for the same duration of the actual run? That would probably not be a sensible default unless the run is reasonably short. 

But if the run is, say, 60 seconds, how do you do the warmup run?

mcollina:
You are right! Making this flexibble would need something like:

```
autocannon -c 100 -d 5 --warmup [ -c 1 -d 3 ] http://localhost:3000
```

You can use https://www.npmjs.com/package/subarg to get the subarg parameters

PR #357, opened 2021-04-07
Feat: Add a warmup option

skywickenden:
### Adds a warmup option so that the server being tested can get up to speed before the stats start to be generated

Closes: https://github.com/mcollina/autocannon/issues/202

### To test:
To test add a warmup flag to the command line or in the config: Eg
```
node autocannon -c 3 -d -W [ -c 2 -d 2 ] http://localhost:5000
```
or
```
node autocannon -c 3 -d --warmup [ --connections 2 --duration 2 ] http://localhost:5000
```
- Check that it works in both command line and imported modes
- Check that additional flags are being correctly utilised and aliased 
- Check that the display on the command line whilst running is displaying the warmup
- Check that the final reported stats do not include the warmup period
- Due to refactoring in `lib/init` check that all calls to `init` work as expected.

### Notes
- Replaces `minimist` with `subarg` so that we can include subarguments such as `--warmup`. `subarg` uses `minimist` internally so the change is minimal - however `subarg` does not use nested aliases and so a new function (`lib/subargAliases`) has been created to enable this
- `lib/init` Has had some refactoring. This was due to the original implementation relying on recursive calling of `_init` which already took considerable cognitive load to trace through the possible paths. Adding a warmup path to the recursion would only have made this situation worse and more bug prone. Instead I have removed the recursive call to `_init` and created two new helper functions that call `_init` exactly as required. 

https://user-images.githubusercontent.com/4930551/114038133-841eae00-9879-11eb-8d1a-a9d2e30732fb.mov

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,10 @@ Available options:
         A path to a Unix Domain Socket or a Windows Named Pipe. A URL is still required in order to send the correct Host header and path.
   -w/--workers
         Number of worker threads to use to fire requests.
+  -W/--warmup
+       Use a warm up interval before starting sampling.
+       This enables startup processes to finish and traffic to normalize before sampling begins
+       use -c and -d sub args e.g. `--warmpup [ -c 1 -d 3 ]`
   --on-port
         Start the command listed after -- on the command line. When it starts listening on a port,
         start sending requests to that port. A URL is still required in order to send requests to
--- a/README.md
+++ b/README.md
@@ -122,6 +122,10 @@ Available options:
   -E/--expectBody EXPECTED
         Ensure the body matches this value. If enabled, mismatches count towards bailout.
         Enabling this option will slow down the load testing.
+  --renderStatusCodes
+        Print status codes and their respective statistics.
+  --debug
+        Print connection errors to stderr.        
   -v/--version
         Print the version number.
   -h/--help
--- a/help.txt
+++ b/help.txt
@@ -17,6 +17,10 @@ Available options:
         A path to a Unix Domain Socket or a Windows Named Pipe. A URL is still required in order to send the correct Host header and path.
   -w/--workers
         Number of worker threads to use to fire requests.
+  -W/--warmup
+       Use a warm up interval before starting sampling.
+       This enables startup processes to finish and traffic to normalize before sampling begins
+       use -c and -d sub args e.g. `--warmpup [ -c 1 -d 3 ]`
   --on-port
         Start the command listed after -- on the command line. When it starts listening on a port,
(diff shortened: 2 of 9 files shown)


README.md
```diff
@@ -118,6 +122,10 @@ Available options:
   -E/--expectBody EXPECTED
         Ensure the body matches this value. If enabled, mismatches count towards bailout.
         Enabling this option will slow down the load testing.
+  --renderStatusCodes
+        Print status codes and their respective statistics.
+  --debug
+        Print connection errors to stderr.
```
skywickenden, line 128: Noticed that these were missing from the readme when adding the warmup flag

simoneb (commented):
Couple of comments but overall LGTM. Clearly reading the code doesn't make it easy to understand if these changes may have introduced regressions. It would be very useful if you:

- could put some inline notes about why you changed some parts of the code
- include testing instructions in the PR description with actions to do and behaviors to expect. A screenshot of how the feature works would also be useful

mcollina (approved):
lgtm

help.txt
```diff
@@ -17,6 +17,10 @@ Available options:
         A path to a Unix Domain Socket or a Windows Named Pipe. A URL is still required in order to send the correct Host header and path.
   -w/--workers
         Number of worker threads to use to fire requests.
+  -W/--warmup
+       Use a warm up interval before starting sampling.
+       This enables startup processes to finish and traffic to normalize before sampling begins
+       use -c and -d sub args e.g. `--warmpup [ -c 1 -d 3 ]`
```
simoneb, line 23: poor pup

```suggestion
       use -c and -d sub args e.g. `--warmup [ -c 1 -d 3 ]`
```

simoneb (commented):
Overall LGTM an works as intended. There's one thing which is unclear though which is why the subargs to -W are required. If it's a by product of using subargs that is fine and we don't have to fight against its default behavior, but if there is an easy way to default them when not provided I believe it would be a great addition because to be honest when you use the new -W option I believe in most cases you just want it to warm up and you don't care about how many connections and duration, 10 for both are sensible defaults and it would be nice if this worked even when they are not provided. In fact, when you do provide at least one of the two, the other gets its default value.

mcollina:
That would be really nice to have @simoneb! Could you make it so @skywickenden.

skywickenden:
@simoneb @mcollina Unfortunately I don't think that can be achieved. `minimist` would require `-W` to be a boolean in all cases to make it parse without a value. Also, if I add a default value then it is used in all cases - even when `-W` is omitted, so a default warmup would always run.

simoneb (approved):

Issue #358, opened 2021-04-07
--forever does not work from the command line


skywickenden:
Using the `-f/--forever` flag on the command line results in an error. EG:

```
node autocannon -f http://localhost:5000
```

results in

```
/Users/sky/NearForm/autocannon/autocannon/autocannon.js:235
    initJob(argv).catch((err) => {
                       ^

TypeError: initJob(...).catch is not a function
    at start (/Users/sky/NearForm/autocannon/autocannon/autocannon.js:235:24)
    at Object.<anonymous> (/Users/sky/NearForm/autocannon/autocannon/autocannon.js:264:3)
    at Module._compile (internal/modules/cjs/loader.js:1156:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1176:10)
    at Module.load (internal/modules/cjs/loader.js:1000:32)
    at Function.Module._load (internal/modules/cjs/loader.js:899:14)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:74:12)
    at internal/main/run_main_module.js:18:47
```

mcollina:
Good spot! Would you like to send a Pull Request to address this issue? Remember to add unit tests.

PR #360, opened 2021-04-13
Fix: forever on command line

skywickenden:
This PR started out to fix the forever bug on the command line. fixes https://github.com/mcollina/autocannon/issues/358

However, it was difficult to unit test the fix using the installed version of tap which required upgrading to tap 15.

tap 15 introduced a new problem, in that the coverage test is stricter than it used to be and as a result total coverage has fallen from the previous 100% to 96% (lines). I have examined all the uncovered code and added additional tests for all the low lying fruit. What remains is difficult to test without refactoring. Much of the current coverage is achieved through functional tests that happen to cover imported functions rather than unit tests and the odd lines not being tested really require unit tests being created for the whole function rather than just the odd fix to create coverage. ~I've added minimum coverage values to `.taprc` to maintain at least the coverage that we currently have.~ I've had to lower the coverage further as node 10 skips some tests and this isn't taken into account in the coverage report
 

Upgrading tap to v 15 also produced a bunch of deprecation warnings for  `t.tearDown`, `t.deepEqual`, `t.is` and `t.false`. These have all been updated to their current variants.

Finally a couple of bugs were found in `validate.js` whilst writing tests for it. I've noted them in the code below.

Screenshot 2021-04-13 at 15 05 20


lib/validate.js
```diff
@@ -11,7 +11,7 @@ const isValidFn = (opt) => (!opt || typeof opt === 'function' || typeof opt ===
 
 const lessThanOneError = (label) => new Error(`${label} can not be less than 1`)
 const greaterThanZeroError = (label) => new Error(`${label} must be greater than 0`)
-const minIfPresent = (val, min) => val && val < min
+const minIfPresent = (val, min) => val !== null && val < min
```
skywickenden, line 14: This use to return `0` if `val` was `0` which resulted in a falsy result when it should be true.

lib/validate.js
```diff
@@ -83,8 +83,15 @@ module.exports = function validateOpts (opts, cbPassedIn) {
...
 
   if (typeof opts.duration === 'string') {
-    if (/[a-zA-Z]/.exec(opts.duration)) opts.duration = timestring(opts.duration)
-    else opts.duration = Number(opts.duration.trim())
+    if (/[a-zA-Z]/.exec(opts.duration)) {
+      try {
+        opts.duration = timestring(opts.duration)
+      } catch (error) {
+        return error
+      }
```
skywickenden, line 91: If `timestring` was invalid then the error was being swallowed. It is now returned as per other errors.

autocannon.js
```diff
@@ -239,9 +239,17 @@ function start (argv) {
...
+      const tracker = initJob(argv)
+      if (tracker.then) {
+        tracker.catch((err) => {
+          console.error(err.message)
+        })
+      }
+    } catch (err) {
       console.error(err.message)
-    })
+    }
```
skywickenden, line 252: The actual fix for `forever`

mcollina (approved):
lgtm

autocannon.js
```diff
@@ -239,9 +239,17 @@ function start (argv) {
       })
     })
   } else {
-    initJob(argv).catch((err) => {
+    // if forever is true then a promise is not returned and we need to try ... catch errors
```
simoneb, line 242: Question: if rather than handling this in a special way we made the forever code path return a promise too, so that we don't have to handle special cases here?

package.json
```diff
@@ -45,7 +45,7 @@
     "sinon": "^10.0.1",
     "split2": "^3.2.2",
     "standard": "^16.0.3",
-    "tap": "^14.10.8"
+    "tap": "^15.0.2"
```
simoneb, line 48: considering that 90% of the changes are tap-related, you could have considered keeping the tap upgrade out of this change and do it standalone instead

simoneb (approved):
LGTM, a couple of comments about the approach though

autocannon.js
skywickenden: This was my original approach but I ran into problems with the reporting. I just tried again now and it is working 😕 . Would you like me to investigate further or shall I merge as is?

simoneb: Up to you. My personal preference would be to do it "properly", but I'm not a maintainer of the repo either so I leave that to you and the maintainers.

simoneb:
bump, can this go in?

Issue #238, opened 2020-01-14
http header values are trimmed instead of used verbatim


sam-github:
It would be useful to be able to specify header values with exact whitespace, like `-H 'field= some value with ows    '`, but they appear to be [trimmed](https://github.com/mcollina/autocannon/blob/7c55bd7d862e62d5090ebe7c9ad7502dfb8b8003/autocannon.js#L145), I'm not sure why.

mcollina:
I think I did it that way because devs are messy. I'm happy to change that if you'd like to send a PR.

elirandav:
can I create a PR for this issue? @mcollina

mcollina:
Go for it!

Issue #170, opened 2018-10-19
Bug, when http header has complex format


gleb-s:
There is an issue, when my content-type is complex and looks like this: `content-type: application/json;odata=verbose`.

So, according to the logics from [this commit](https://github.com/phra/autocannon/commit/532349fb2318becacf73041604f6dea91394b36d#diff-b8cd744f56f8425866b8960637fe3f3bR96), I suppose, that my server is getting header like this:
`content-type: application/json;odata: verbose` - this syntax is incorrect.
So it would be nice, if someone could fix it.

_Originally posted by @gleb-s in https://github.com/mcollina/autocannon/issues/133#issuecomment-431030961_

phra:
hi @gleb-s 

according to my commit only the first occurrence of `=` is matched using [indexOf](https://developer.mozilla.org/it/docs/Web/JavaScript/Reference/Global_Objects/Array/indexOf) so I'm not sure about your supposition.

phra:
hi @gleb-s 

i made https://github.com/mcollina/autocannon/pull/171 in order to fix your issue.
i have also added the required unit tests to prevent this from happening in the future.
thanks for reporting it.

/cc @mcollina

PR #365, opened 2021-05-05
fix(argument parsing): remove trim of headers

elirandav:
BREAKING CHANGE: clients expecting the header to be trimmed will the get
wrong header

Closes https://github.com/mcollina/autocannon/issues/238


elirandav:
can you please review and comment? @mcollina

mcollina:
@phra could you take a look? You edited those files before and you added the trim(), however it didn't seem it was needed to fix the original issue of https://github.com/mcollina/autocannon/pull/170 and https://github.com/mcollina/autocannon/pull/171

mcollina (approved):
lgtm

Issue #361, opened 2021-04-13
Pass headers in "onResponse" of requests array


Elias-Serneels:
I'm trying to test an application which has the server set a cookie in the response, redirects the user to a different path and then checks the for the cookie in the second request. 
So in short:
In step 1 I send a GET to the server which responds with a 302 and a cookie.
In step 2 I need go to the location of the 302 and send along the cookie I received in the first request.

I'm using the autocannon requests array to set up the sequence of requests that need to happen.

My problem is that the "onResponse" function doesn't expose the headers that are returned from the first response. This means I can't access the "set-cookie" header which in turn means I can't provide the cookie in the header of the next request

My autocannon setup is like this:
``` import * as autocannon from "autocannon";

autocannon({
    url: "localhost:3000",
    requests: [
        {
            // Start the request
            method: 'GET',
            path: '/auth,
            onResponse: (status, body, context) => {
            // here I would like to access the "set-cookie" header so I can add it to the context like this:
            // context.cookie = headers["set-cookie"];
            // but no headers are passed to the onResponse method.
                context.path = body.url;
            }
        },
        {
            // Now go the path returned in the previous response and send along the cookie
            method: 'GET',
            setupRequest: (req: any, context: any) => ({
                ...req,
                path: context.path,
                headers: {"cookie": context.cookie}
            }),
            onResponse: (status: number, body: string, context: { [key: string]: any }, headers: any) => {

            }
        }
```

Are there any ways around this? Is there another way I can use autocannon to achieve what I need?

I checked out the clients API but I don't think the client.on("headers") event helps my case as it doesn't keep track of where in the request cycle I am. Further it just gives the headers keys and values back as a flat list so I have no idea of knowing where a header stops and the next header begins.

A solution for me would be to pass on the headers in the onResponse method. I can make a PR for this but I'd like to know if there are alternatives first.

mcollina:
I would recommend you to go ahead with a PR, thanks!

EladIsraeli:
I would like to take this PR :)

mcollina:
Go for it!

Elias-Serneels:
Thanks for picking up this PR! I just got back to what I was working on back then and saw I never went ahead and made the PR with the changes. Good to see it merged already!

Elias-Serneels:
After testing this out it works great in most cases however it doesn't seem to work when a server sends back multiple headers with the same key but a different value. For example my server sends back multiple `set-cookie` headers.

[...]

PR #366, opened 2021-05-05
feat: added headers to the onResponse callback.

EladIsraeli:


EladIsraeli:
Resolves https://github.com/mcollina/autocannon/issues/361

mcollina (commented):
Could you add docs?

EladIsraeli:
Yeah sure, I added docs let me know if there is anything related to the docs needed here

lib/requestIterator.js
```diff
@@ -106,9 +106,9 @@ RequestIterator.prototype.rebuildRequest = function () {
   return data
 }
 
-RequestIterator.prototype.recordBody = function (request, status, body) {
+RequestIterator.prototype.recordBody = function (request, status, body, headers) {
   if (request && typeof request.onResponse === 'function') {
-    request.onResponse(status, body, this.context)
+    request.onResponse(status, body, this.context, headers)
```
mcollina, line 111: Can you move the array-to-object logic within this if? We can do it only if `onResponse` is set.

mcollina (changes requested):

lib/requestIterator.js
EladIsraeli: Sounds good! it could spare us an unnecessary conversation.

lib/requestIterator.js
```diff
@@ -106,9 +106,19 @@ RequestIterator.prototype.rebuildRequest = function () {
...
 }
 
-RequestIterator.prototype.recordBody = function (request, status, body) {
+RequestIterator.prototype.recordBody = function (request, status, body, headers) {
   if (request && typeof request.onResponse === 'function') {
-    request.onResponse(status, body, this.context)
+    request.onResponse(status, body, this.context, toHeaderKeyValue(headers || []))
+  }
+
+  function toHeaderKeyValue (headersArray) {
```
mcollina, line 114: Can you move this to a top level function?

lib/requestIterator.js
```diff
@@ -106,9 +106,19 @@ RequestIterator.prototype.rebuildRequest = function () {
...
-    request.onResponse(status, body, this.context)
+    const toHeaderKeyValue = (headersArray) => {
+      const headersKeyValue = {}
+
+      for (let i = 0; i < headersArray.length - 1; ++i) {
+        headersKeyValue[headersArray[i]] = headersArray[i + 1]
+      }
+
+      return headersKeyValue
+    }
```
mcollina, line 119: by top level I mean to not allocate a closure and call a function defined outside of this one.

Issue #320, opened 2021-01-12
Provision to specify urlSafe when idReplacement is true


zebkailash:
If we want to add an id after an email string eg: abc_randomStr@gmail.com but currently there is no way to create a url-safe id.

Currently, it creates a string **kailash-FZXY40aXS5C6Ph0KXs1KSQ/0000008544@gmail.com** which is an invalid email due to slash. Can this be supported in upcoming releases?

mcollina:
> Can this be supported in upcoming releases?

It will if you send a PR!

PR #367, opened 2021-05-05
fix Provision to specify urlSafe when idReplacement is true #320

namerhila:
fix issue Provision to specify urlSafe when idReplacement is true #320 https://github.com/mcollina/autocannon/issues/320


mcollina (changes requested):
Can you please add a unit test?

Issue #356, opened 2021-04-06
Input is not working for me


shlomi-ozon:
Hello,
I am trying to use the input (-i) option and passing the relative path to a text file.
The content of the file is a stringify of an object.
When I am trying to run it I get "Worker error: Error: body must be either a string or a buffer".
so I get into the code to check how it works and I find out that in function "parseArguments" the content of the body after fetching the file is a buffer (like it should be when you read with readFileSync) but when the data is reaching the "httpRequestBuilder" the body is not a buffer anymore.
I tried it even with a simple string inside and I got the same result.
any idea why it works like that? am I doing something wrong?

my call:
autocannon -d 10 -a 1 -w 5 -i ./message.txt -H requestid=u9foctpc8ugq9fye4b3c766gln21tny796dzecwxcwntr1ygqkunmlgc9bpz6zk0 -H Content-Type=application/json  -m POST http://localhost:3000/call

shlomi-ozon:
Ok, when I am running it without the workers it is working fine.
So I guess the problem it there.

mcollina:
Good spot! Would you like to send a Pull Request to address this issue? Remember to add unit tests.

shlomi-ozon:
I will take a look later, I am not sure I will have spare time to fix it.
if it something that won't take too long I will do it gladly

elirandav:
can I create a PR for this issue? @mcollina

shlomi-ozon:
@elirandav I already created the fix, but I can't create tests for it currently. do you want I will posh my fix and you will add a test for it?

elirandav:
nice @shlomi-ozon. Actually, I work TDD so it will be challenging to do the opposite but I see my fix PR was already approved.

PR #368, opened 2021-05-05
fix(multi workers and input file): working with multiple workers and…

elirandav:
… input file

Closes https://github.com/mcollina/autocannon/issues/356


mcollina (approved):
lgtm

PR #400, opened 2021-10-13
Replaced 'clone' library with 'lodash.clonedeep'

### LINKED ISSUE #391: Unable to load PFX certificate with tlsOptions + workers
Example config -

`const instance = autocannon({
url: 'https://localhost:8080',
connection: 10,
workers: 3,
tlsOptions: {
pfx: fs.readFileSync('path/of/p12/keystore')
passphrase: 'test'
}
}`

When I use this config. I get 'Error: Unable to load PFX certificate'

This config works when I don't use 'workers' option.

radomird:
Replaced the [clone](https://github.com/pvorb/clone) library with [lodash.clonedeep](https://www.npmjs.com/package/lodash.clonedeep). Main reason is this [issue](https://github.com/pvorb/clone/issues/69) on the `clone` library.

Closes #391


simoneb (commented):
Overall looks good. So this solution works without using the v8 serialization APIs that Matteo suggested, right? Let's make sure the test you wrote consistently fail without the production code change.

test/workers.test.js
```diff
@@ -175,7 +175,7 @@ test('tlsOptions using pfx work as intended in workers', { skip: !hasWorkerSuppo
     }
   }, function (err, result) {
     t.error(err)
-    t.ok(result.requests, 'requests are ok')
+    t.ok(result, 'requests are ok')
```
simoneb, line 178: does result contain something that suggests that the run succeeded?

radomird: Yes, if the result is undefined that means it was **not** successful. This particular issue fails with result being undefined, but I've just realised that we also check for existence of an error `t.error(err)` - I think that would be enough.
I'll remove the unnecessary check for the result.

simoneb: you can keep it if that's part of the API that a non-null result means success

radomird: Okay, then no updates are necessary. That was, indeed, my intention in the first place..

PR #395, opened 2021-09-16
docs: grammar, spelling, and readability fixes

Fdawgs:
Also closes #363

Issue #363, opened 2021-05-04
Why 4.78m and 5 seconds in sample of readme?


Giancarlo-Ma:
I think that should be  2.29 MB and 10 seconds.

mcollina:
Would you like to send a PR to update?

PR #404, opened 2021-11-10
Update: example values in readme

Jdiego06:
**2.29MB** correspond to the amount of data dowloaded in **1 second** in the worst case, while the entire test took **10 seconds** to complete.

Related with #395 and #363


mcollina (approved):
lgtm

Issue #413, opened 2021-12-15
Can't send JSON array as body


silverwind:
```shell
$ autocannon -b '[]' https://localhost
body must be either a string or a buffer
```

Adding some logging, it seems this is an issue with the argument parser which decodes an object from the argument, `body` is `{"_":[]}` when it hits this verification function.

Ugly workaround:

```js
    } else if (body && Array.isArray(body._)) {
      bodyBuf = Buffer.from(`[${body._}]`);
```

mcollina:
good spot! Would you like to send a Pull Request to address this issue? Remember to add unit tests.

silverwind:
Possibly. I guess the [`subarg`](https://github.com/substack/subarg) is the culprit with its interpretion of `[` in arguments. It probably needs to be either replaced with `minimist` or above dirty workaround to be added.

mcollina:
I'm kind of ok with the ugly workaround if it fixes your problem.

PR #414, opened 2021-12-16
fix: Fix JSON array body on CLI

silverwind:
Fixes: https://github.com/mcollina/autocannon/issues/413


mcollina (approved):
lgtm

silverwind:
By the way: I did not explore the option to replace `subarg` with `minimist` (which I think is the proper solution), because I saw `subarg`'s context feature being in active use, so the replacement would be a breaking one (thought it's one I'd be willing to take).

PR #423, opened 2022-01-11
Add support for multi-value headers in onResponse

Elias-Serneels:
Fixes the problem I have with multi-value headers (such as "set-cookie") that I explain in my comment in #361


mcollina (approved):
lgtm

PR #425, opened 2022-01-30
Feature make samples variable

drewva:
Hopefully this actually works because I didn't test it with every configuration since I figured you all probably have that automated. I did play around with some configurations though.


mcollina (changes requested):
Could you add a unit test?

drewva:
Would you rather I add unit tests to the existing testing files & then commit them, or should I provide a single, new file that tests all of the things I added?

mcollina (approved):
lgtm

mcollina (approved):
lgtm

mcollina:
linting is failing, could you take a look?

mcollina (approved):
lgtm

PR #435, opened 2022-03-15
Fixes #432 - Updating to only load the help text when required

### LINKED ISSUE #432: Cannot Use on OpenWhisk Actions Using Webpack
I was trying to use autocannon in an OpenWhisk action created with webpack. It seems like when webpack packages up autocannon it doesn't include the dynamic call to read help.txt thus autocannon fails when invoked with:

    Initialization has failed due to: Error: ENOENT: no such file or directory, open '/nodejsAction/G7zjExia/help.txt'
    at Object.openSync (fs.js:476:3)
    at Object.readFileSync (fs.js:377:35)
    at Object.52430 (/nodejsAction/G7zjExia/index.js:79610:17)
    at __webpack_require__ (/nodejsAction/G7zjExia/index.js:186808:42)
    at Object.78163 (/nodejsAction/G7zjExia/index.js:24:20)
    at __webpack_require__ (/nodejsAction/G7zjExia/index.js:186808:42)
    at /nodejsAction/G7zjExia/index.js:186875:37
    at Object.<anonymous> (/nodejsAction/G7zjExia/index.js:186878:12)
    at Module._compile (internal/modules/cjs/loader.js:1063:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1092:10)

My thought would be to _either_ wrap the read of help.txt in a try-catch or convert the help.txt into a JSON that can be included with require and thus be found by webpack when it shakes the tree.

klcodanr:
Updated to move the call to load the help text inside the `parseArguments` which only gets called when autocannon is invoked via the CLI. For module imports, the default export is the `initJob` function which does not call parseArguments. 

You could still run into this problem if you tried to import the parseArguments function from inside a webpack / otherwise packaged module, e.g.:

    const { parseArguments, start } = require('autocannon');
    const args = // some source of arguments
    start(parseArguments(args)); 

...but honestly I'd see that as an edge case.

### DIFF
--- a/autocannon.js
+++ b/autocannon.js
@@ -12,7 +12,6 @@ const spawn = require('child_process').spawn
 const managePath = require('manage-path')
 const hasAsyncHooks = require('has-async-hooks')
 const subarg = require('subarg')
-const help = fs.readFileSync(path.join(__dirname, 'help.txt'), 'utf8')
 const printResult = require('./lib/printResult')
 const initJob = require('./lib/init')
 const track = require('./lib/progressTracker')
@@ -119,6 +118,7 @@ function parseArguments (argvs) {
   }
 
   if (!checkURL(argv.url) || argv.help) {
+    const help = fs.readFileSync(path.join(__dirname, 'help.txt'), 'utf8')
     console.error(help)
     return
   }


mcollina (approved):
lgtm

mcollina:
Thanks for bearing with me!

klcodanr:
No problem at all! Thanks for being so responsive!

PR #437, opened 2022-03-22
Add cli options to pass cert information

chiragpat:
### Why
Currently the only way to pass custom cert information to autocannon is via the `tlsOptions` in the autocanno js api. This PR adds the ability to pass in common cert information from the cli tool so that consumers can easily use autocannon without writing a custom script.

### How
- Add logic in autocannon.js to parse arguments and read the relevant files
- Normalize all the arguments to be part of the `tlsOptions` key so that the `httpClient.js` can get the correct values

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,8 +124,14 @@ Available options:
         Enabling this option will slow down the load testing.
   --renderStatusCodes
         Print status codes and their respective statistics.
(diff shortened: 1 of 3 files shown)


mcollina (approved):
lgtm

PR #454, opened 2022-07-11
ci!: update workflow

Fdawgs:
BREAKING CHANGE: Drops support for EOL node 12 and 15; adds support for node 18

This PR:
-   Drops support for EOL node versions
-   Removes Git credentials/SSH keys after checkout as a security precaution by setting `persist-credentials: false`, they are not used after the initial checkout
-   Enables `concurrency`; see [related docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency), this allows a subsequently queued workflow run to interrupt previous runs in PRs
-	Adds a conditional `if` to the automerge job; this stops the job from running if the user is not Dependabot, saving a few seconds CI run time

closes #427 
closes #431
closes #456


mcollina:
@Fdawgs apparently this fails on CI on Node v18. Could you take a look?

Fdawgs:
Not 100% on this one. Seems to be an issue with passing `pfx` to [tls.createSecureContext()](https://nodejs.org/docs/latest-v18.x/api/tls.html#tlscreatesecurecontextoptions). Swapped it out with a generated `.pfx` file and it throws an error as well:

image


I've looked through the V17 and V18 changelogs and can't spot any breaking changes in this area. 😢

mcollina:
Maybe skips?

Fdawgs:
> Maybe skips?

Found the issue: https://github.com/nodejs/node/issues/40672

Have recreated the test pkcs12 file using OpenSSL v3 and it now uses a supported cipher.

Fdawgs:
Don't know if it's worth adding `chalk` and `pretty-bytes` to the list of dependencies to not automerge, considering they're now fully ESM?

mcollina:
> Don't know if it's worth adding chalk and pretty-bytes to the list of dependencies to not automerge, considering they're now fully ESM?

add them!

Fdawgs:
> add them!

Cool beans, all done @mcollina

mcollina (approved):
lgtm

mcollina:
CI is failing :(

Fdawgs:
> CI is failing :(

There we go!

mcollina:
thx

PR #461, opened 2022-09-05
fix verifyBody documentation

madflow:
* The docs in the readme seemed to be a copy+paste from the `setupClient` function in the preceding line. 
* `verifyBody` actually accepts the response body ("passed in")
* Mention "mismatches"


mcollina (approved):
lgtm

PR #463, opened 2022-09-20
Accept host header case-insensitively

silverwind:
This makes CLI `-H 'Host: foo'` work where previously it did not identify the header correctly because it was case-sensitive.

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -13,6 +13,7 @@ jobs:
     permissions:
       contents: read
     strategy:
+      fail-fast: false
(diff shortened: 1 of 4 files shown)


silverwind:
Will try to fix the test, I didn't test locally because I see around 30 test failures even on a clean branch.

silverwind:
Should be good to go now. Previous failures were odd timing flakes on the CI. I also disabled `fail-fast` behaviour because I think it's much more useful to have all jobs to always run to completion.

mcollina (approved):
lgtm

PR #469, opened 2022-10-15
doc: change recomendation to wrk2 instead wrk

RafaelGSS:
`wrk` doesn't handle coordinate omission. Furthermore, autocannon is based on `wrk2`.


mcollina (approved):
lgtm

PR #483, opened 2023-03-12
Fixed #482

### LINKED ISSUE #482: Forever option fails from second iteration onwards
It seem like when you run with `forever` option the first iteration works perfectly fine but the next ones just gets sort-circuited and cancelled cause of [this line](https://github.com/mcollina/autocannon/blob/b198514dc02105f8b93a2e7c7940f708a50876da/lib/run.js#L155) which in my debugging opinion should be like [this line](https://github.com/mcollina/autocannon/blob/b198514dc02105f8b93a2e7c7940f708a50876da/lib/run.js#LL88C11-L88C11) which executes all iteration correctly....

again I am no expert on this repo, so excuse me if I am wrong

LRagji:
All test passing!!

PS: Serial test is failing even before this fix, if someone can look into that.

### DIFF
--- a/lib/run.js
+++ b/lib/run.js
@@ -152,9 +152,11 @@ function run (opts, tracker, cb) {
 
       const restartFn = () => {
         stop = false
-        stopTimer = setTimeout(() => {
-          stop = true
-        }, opts.duration * 1000)
+        if (!amount) {
+          stopTimer = setTimeout(() => {
+            stop = true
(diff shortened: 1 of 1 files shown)


mcollina (approved):
lgtm

mcollina (changes requested):
Could you add a test for the fix?

PR #486, opened 2023-04-24
fix Cannot access 'connection' before initialization

moty66:
trying with node version v16.14.2 I get this error 

```
node samples/customise-individual-connection.js 
/Users/.../workspace/autocannon/samples/customise-individual-connection.js:27
    client.setBody('connection number', connection++)
           ^

ReferenceError: Cannot access 'connection' before initialization
... 
... 

```


mcollina (approved):
lgtm

PR #488, opened 2023-04-25
Defer aggregating results until after the instance returns to allow multi-instance load tests

milesnash:
This update introduces the ability to coordinate a load test using multiple separate machines running autocannon against an API, whilst still being able to use the built-in result aggregation functionality, as if it were using connections and/or worker threads on a single machine.  It is a lightweight and backward-compatible.

To keep things simple in autocannon, the implementation of the coordination between machines is left as the responsibility of the consumers of this library, as there are undoubtedly many possibilities.


mcollina (approved):
lgtm

PR #491, opened 2023-07-02
Add Node 20 to the test matrix

kibertoad:


kibertoad:
Curious to see if this passes, locally autocannon is failing on Node 20 for me, wonder if it's me or if it's autocannon.

PR #496, opened 2023-09-15
perf: use `node:` prefix to bypass require.cache call for builtins

Fdawgs:
Introduced as part of v16.0.0 and  v14.18.0, this allows redundant `require.cache` calls to be bypassed for builtin modules, saving a few yoctoseconds.

See https://nodejs.org/api/modules.html#core-modules and [discussion in nodejs/node repo regarding why `require.cache` calls are redundant for builtins](https://github.com/nodejs/node/pull/37246/files#r588397158).

### DIFF
--- a/samples/bench-multi-url.js
+++ b/samples/bench-multi-url.js
@@ -1,6 +1,6 @@
 'use strict'
 
-const http = require('http')
+const http = require('node:http')
 const autocannon = require('../autocannon')
 
 function createHandler (serverName) {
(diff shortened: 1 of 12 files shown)


Fdawgs:
Bit stumped on the test failure. :|

Fdawgs:
Looks like tests are unhappy prior to this change anyway. Just updated the samples https://github.com/mcollina/autocannon/pull/496/commits/54c404fb7d949e57156e4aa3e5b0cac864b1d536, which have no bearing on anything, and the CI fails.

PR #500, opened 2023-11-22
lib: add verbose option

RafaelGSS:
This PR adds a new option to the CLI `--V` to manage the verbosity of the output. Sometimes (mostly for automation) we just need the last line output:

```console
$ ./autocannon.js -V=0 -c 100 -d 30 localhost:3000
Running 30s test @ http://localhost:3000
100 connections

44k requests in 30.02s, 8.06 MB read
```

By default verbose is `true` so nothing will change to regular users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,6 +136,8 @@ Available options:
         Print connection errors to stderr.
   -v/--version
         Print the version number.
+  -V/--verbose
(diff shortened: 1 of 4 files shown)


mcollina (approved):
lgtm

PR #515, opened 2024-03-12
fix(requestIterator): generated IDs were not URL-safe

### LINKED ISSUE #320: Provision to specify urlSafe when idReplacement is true

10xLaCroixDrinker:
Fixes #320

### DIFF
--- a/lib/requestIterator.js
+++ b/lib/requestIterator.js
@@ -1,6 +1,6 @@
 'use strict'
 
-const hyperid = require('hyperid')(true)
+const hyperid = require('hyperid')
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

PR #521, opened 2024-04-10
Allow for ID replacement in the URL

### LINKED ISSUE #512: Is there a way to generate a url for each request?
I'd like to benchmark `POST` requests where I could pass a random id at the end of path.
I tried using `setupClient` to change the url, but I'll only change the first request of each client.

MattIPv4:
Resolves #512

### DIFF
--- a/lib/requestIterator.js
+++ b/lib/requestIterator.js
@@ -108,8 +108,12 @@ RequestIterator.prototype.rebuildRequest = function () {
     this.currentRequest.headers = this.currentRequest.headers || this.headers
     data = this.requestBuilder(this.currentRequest, this.context)
     if (data) {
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

MattIPv4:
👋 @mcollina it looks like a single test job failed here, which prevented automerge. Could you trigger a re-run of the failed job?

GlenTiki:
That test has failed on rerun twice more. Three fails seems unlikely to be flakey tests - this might be an edgecase on node 20 on windows. Possibly in node?

GlenTiki:
That test is failing on main, anyway. Merging since its not a problem with your code.

Issue #180, opened 2018-12-01
assume --on-port if there is '-- node server.js'


mcollina:
Currently we have to do:

```
autocannon  --on-port / -- node bare
```

However it would be more usable/friendly if we did

```
autocannon  / -- node bare
```

We can basically assume `--on-port` if there is `-- node`.

PR #530, opened 2024-10-11
assume onPort if -- node is set

karankraina:
Adds support for #180 

@mcollina Please review when you have some time.

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -16,7 +16,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
-        node-version: [14, 16, 18, 20]
+        node-version: [18, 20, 22]
 
     steps:
     - uses: actions/checkout@v3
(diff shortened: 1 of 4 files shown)


mcollina (changes requested):
I think we can just remove `--onPort`.

Remember to adjust the docs.

karankraina:
Thanks for your review @mcollina . Also, if we remove onPort flag, I think the check should not striclty be restricted to `-- node`. We should assume on Port if `-- <any-script>` is provided. Your thoughts ?

mcollina:
Ah, that's true. Then, this is good. Can you just update the docs?

karankraina:
Sure, updating.

karankraina:
@mcollina Does it look okay to you?

mcollina (approved):
lgtm

karankraina:
@mcollina checks on node v14 on mac and node v20 on windows are failing. Can see other merges with these fails too. Any known issues ?

mcollina:
Drop node 14 and v16 support, I'll bump
the major.

windows is likely flaky.

karankraina:
Should I add v22 ?

mcollina:
yes please

karankraina:
Done but windows still not happy. I think we need another issue to address that. And if this looks good, can we merge it?

karankraina:
@mcollina

mcollina:
I'll look into the windows tests asap.

mcollina (approved):
lgtm

PR #535, opened 2025-02-24
ci: update github actions

Fdawgs:
This PR bumps GitHub Actions to their latest major versions. Main change is they are now using Node 20 instead of 16, which was EOL as of April 2022:
   - Changelog for `actions/checkout` can be [found here](https://github.com/actions/checkout/blob/main/CHANGELOG.md)
   - Releases for `actions/setup-node` can be [found here](https://github.com/actions/setup-node/releases)


mcollina:
why did you close this?

Issue #528, opened 2024-08-01
Replace third-party form-data package with native FormData class


JaoodxD:
Node.js has supported native `FormData` since version 18. 
This issue proposes replacing the third-party `form-data` package with the built-in `FormData` class to reduce number of third-party dependencies and lean toward native solutions.

mcollina:
Thanks for reporting! Would you like to send a Pull Request to address this issue? Remember to add unit tests.

JaoodxD:
Sure, I'd be happy to work on this

gusanthiago:
Hello @mcollina @JaoodxD, I sent a PR to close this issue, can you check please ?

Thanks :)

PR #541, opened 2025-09-03
chore(lib): Replace form data third-party to nodejs native form-data

gusanthiago:
This PR aim replace third-party form-data package with native FormData as describe 

Related to https://github.com/mcollina/autocannon/issues/528

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,9 +6,6 @@
   "bin": {
     "autocannon": "autocannon.js"
   },
-  "engines": {
-    "node": "^18.20.8"
-  },
   "scripts": {
     "test": "standard && tap test/serial/*.test.js test/*.test.js",
     "standard:fix": "standard --fix"
--- a/package.json
+++ b/package.json
@@ -6,6 +6,9 @@
   "bin": {
     "autocannon": "autocannon.js"
   },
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -6,6 +6,9 @@
   "bin": {
     "autocannon": "autocannon.js"
   },
+  "engines": {
+    "node": "^18.20.8"
+  },
```
mcollina: please remove this block

mcollina (changes requested):
Why can't we use the actual FormData object?

gusanthiago:
@mcollina are you saying inside the return object? FormData doesn't have the getBuffer that is used in https://github.com/mcollina/autocannon/blob/master/lib/run.js#L65, so it's a way to adapt the interface of that.

Wdyt about it ?

About tests, looks like a broken test are failing

mcollina:
https://github.com/mcollina/autocannon/issues/528 asked to change from form-data to the native FormData object in Node.js. This PR does not implement that.

As you are saying, this might not be possible, so we could keep using form-data.

gusanthiago:
Make sense, I gonna close this PR, tks @mcollina

PR #544, opened 2026-04-28
fix: emit trailers event for HTTP response trailers

mcollina:
## Summary
- The `client supports response trailers` test in `test/httpClient.test.js` has been silently broken since #49 (2016): that PR added the test alongside `kOnHeaders = () => {}`, which discards trailers. With Node's old built-in `http-parser`, the `body` callback included raw chunked-encoding bytes (frame headers + payload + trailers concatenated), so the test's regex on `raw.toString()` happened to match. With `http-parser-js@0.5.10`, `body` only contains the decoded payload — trailer bytes are delivered to `kOnHeaders` after `kOnHeadersComplete`.
- Replace the no-op `kOnHeaders` handler with one that emits a `trailers` event carrying the raw `[name, value, ...]` array.
- Update the test to listen for `trailers` and assert the expected `Content-MD5` value, so it actually exercises trailer handling.

## Test plan
- [x] `npx tap test/httpClient.test.js` — 46/46 pass on Node 22
- [ ] CI green on ubuntu/windows/macOS × Node 20/22


lraveri-talentware (approved):
I was investigating the CI failure and reached the same conclusion

